PR #73, opened 2014-10-18
Fixing Purchase Free Items

einsteinK:
Currently the only thing that goes wrong is, when pressing the purchase-button, it waits a second and displays:
"Roblox is currently in maintenance blablabla"     Fiddler shows an 'Internal Server Error'
Apparently, the HttpRequest fails internally when the 'purchasePrice'-paramter is "nil", as was the case for free items.
With this fix, it'll set the parameter to '0', which is accepted.
(Tested with Fiddler, spoofed the purchase request from the client)

This should be the last time we need to change something to fix the purchase of Free Items.


matthewdean:
What will it take to get this merged? Many forum users have been asking for this feature.

hockeychaos:
sorry, this one fell off the radar a bit.  It was easier for me to make a new change rather than merge, but the fix is there now.

PR #95, opened 2015-01-15
Chat ignore feature

SolarCrane:


SolarCrane:
Merging in the Chat Ignore Feature.

PR #101, opened 2015-01-21
US26136: Update in game notifications

jmargh:
Updating the in game notification system. This implements the notification system in lua and updates all core scripts that were using a GuiService:SendNotification() call. All new functionality is wrapped in fast flags.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/NotificationScript2.lua
+++ b/CoreScriptsRoot/CoreScripts/NotificationScript2.lua
@@ -1,0 +1,472 @@
+--[[
+		Filename: NotificationScript2.lua
+		Written by: jmargh
+		Description: Handles notification gui for roblox in game events
+--]]
+
+--[[ Services ]]--
+local BadgeService = game:GetService('BadgeService')
+local GuiService = game:GetService('GuiService')
+local Players = game:GetService('Players')
+local PointsService = game:GetService('PointsService')
+local MarketplaceService = game:GetService('MarketplaceService')
+local TeleportService = game:GetService('TeleportService')
+local Settings = UserSettings()
+local GameSettings = Settings.GameSettings
+
+--[[ Script Variables ]]--
+local LocalPlayer = nil
+while not Players.LocalPlayer do
+	wait()
+end
+LocalPlayer = Players.LocalPlayer
+local RbxGui = script.Parent
+local NotificationQueue = {}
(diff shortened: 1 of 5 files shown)


jmargh:
Dan doesn't like this.

CoreScriptsRoot/CoreScripts/NotificationScript2.lua
```diff
@@ -0,0 +1,472 @@
...
+local EASE_STYLE = Enum.EasingStyle.Sine
+local TWEEN_TIME = 0.35
+local DEFAULT_NOTIFICATION_DURATION = 5
+
+--[[ Images ]]--
+local PLAYER_POINTS_IMG = 'http://www.roblox.com/asset?id=155363793'
+local BADGE_IMG = 'http://www.roblox.com/asset?id=177200377'
+local FRIEND_IMAGE = 'http://www.roblox.com/thumbs/avatar.ashx?userId='
+local GRAPHICS_QUALITY_LEVELS = 10
+local GRAPHICS_QUALITY_TO_INT = {
```
einsteinK: Why not just use the Enum's .Value field?
For SavedQualitySetting it'll return the same value as predicted by this table.

CoreScriptsRoot/CoreScripts/NotificationScript2.lua
```diff
@@ -0,0 +1,472 @@
...
+	end
+end
+LocalPlayer.FriendStatusChanged:connect(onFriendStatusChanged)
+
+--[[ Player Points Notifications ]]--
+local time = 20
+local function onPointsAwarded(userId, pointsAwarded, userBalanceInGame, userTotalBalance)
+	--if userId == LocalPlayer.userId then
+		if pointsAwarded == 1 then
+			sendNotifcation("Point Awarded", "You received "..tostring(time).." point!", PLAYER_POINTS_IMG, DEFAULT_NOTIFICATION_DURATION, nil)
```
einsteinK: tostring(time) ?
Shouldn't it be "You received "..pointsAwarded.." point!" etc?
The local *time* seems to be useless...
(Except the tostring-case, it isn't used anywhere)

jmargh: Yep, it was left in from a test case I was running, to test variable time notifications. Will be updated.

PR #102, opened 2015-01-21
US26136: Update in game notifications.

jmargh:
Implements a new notification system for in game notifications (badges, player points, friend request, etc). This replaces and updates the engine notification system, porting it to lua. Updated all core scripts that used the old notification system. All new functionality is wrapped in fast flags.


jmargh:
Closing to make some fixes and put in new icons from the art team. Will resubmit later.

PR #103, opened 2015-01-21
US26136: Update in game notifications.

jmargh:
This updates the in game notification system, porting it to lua. It implements it with the new gui style, with some changes to how notifications are displayed. All new features are wrapped in fast flags.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/NotificationScript2.lua
+++ b/CoreScriptsRoot/CoreScripts/NotificationScript2.lua
@@ -1,0 +1,459 @@
+--[[
+		Filename: NotificationScript2.lua
+		Written by: jmargh
+		Description: Handles notification gui for the following in game ROBLOX events
+			Badge Awarded
+			Player Points Awarded
+			Friend Request Recieved/New Friend
+			Graphics Quality Changed
+			Teleport Failed (studio only)
+			CreatePlaceInPlayerInventoryAsync
+--]]
+
+--[[ Services ]]--
+local BadgeService = game:GetService('BadgeService')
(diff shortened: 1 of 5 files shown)


einsteinK:
Nothing important, but it seems when unchecking the "Auto" option for VideoQuality,
there could be a notification appearing.
(Due the new level then being > 0, allowing for a notification)
Seems to differentiate from the old system, which used some weird magic to counter this.

CoreScriptsRoot/CoreScripts/NotificationScript2.lua
```diff
@@ -0,0 +1,459 @@
...
+	end
+end
+BadgeService.BadgeAwarded:connect(onBadgeAwarded)
+
+--[[ Graphics Changes Notification ]]--
+GameSettings.Changed:connect(function(property)
+	if property == "SavedQualityLevel" then
+		local level = GameSettings.SavedQualityLevel.Value
+		-- value of 0 is automatic, we do not want to send a notification in that case
+		if level > 0 and level ~= CurrentGraphicsQualityLevel then
```
jmargh: The case for SavedQualityLevel being 0 is handled here.

jmargh:
https://github.com/ROBLOX/Core-Scripts/blob/master/CoreScriptsRoot/CoreScripts/NotificationScript2.lua#L343

This handles both the cases if it's set to auto and coming back to manual.
Also, thanks for pointing out the previous issues =)

einsteinK:
Ah, I totally ignored the fact that CurrentGraphicsQualityLevel doesn't get set for Auto.

I was bored, and just felt like checking the new script out.

PR #106, opened 2015-01-27
Update BackpackScript.lua

Scarfacial:
- Set HotbarFrame.Active to false
HotbarFrame currently blocks a large portion of screen space at the bottom because it is 'Active', but when not active it functions the same and doesn't block lower elements from being interacted with.


Only20:
Different solution preferred, as seen here: f54d5cbe2b7be5fc272195855e9d53f19224f6fa

PR #114, opened 2015-02-12
Added mouse sensitivity to rootcamera and improved efficiency of the cameras.

CodedNil:
RootCamera.Sensitivity can eventually be turned into some easily accessibly value of player or mouse

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/PlayerScripts/CameraScript.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/PlayerScripts/CameraScript.rbxmx
@@ -6,8 +6,7 @@
 			<bool name="Disabled">false</bool>
 			<Content name="LinkedSource"><null></null></Content>
 			<string name="Name">CameraScript</string>
-			<ProtectedString name="Source">local RunService = game:GetService(&apos;RunService&apos;)
-local UserInputService = game:GetService(&apos;UserInputService&apos;)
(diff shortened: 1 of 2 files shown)


SolarCrane:
Mouse sensitivity sounds like a good idea to eventually have in the settings or read from the OS, but I am not confident in this implementation. It looks like one of big things that you are doing is replacing the renderstepped update with a mouse moved update loop. Unfortunately, there are a very large number of cases where the camera needs to be updated even if the mouse has not received updates.

Examples may include:
Moving your character with the keyboards
An object moving in-between you and the camera and the camera needs to pop in
Updating character transparency based on the distance from your character changing
Zooming in and out or moving the camera around using keyboard commands (i,o, <, >, pageup, pagedown)
Values changing such as CameraOffset on humanoid
The CameraSubject changed to a new subject

CodedNil:
Ah, didnt realise about those flaws with the mouse changed instead of render stepped, probally due to my testing enviroment having the coregui camera running at the same time also running the poppercam to make up for when my camera didnt use it.
And with mouse sensitivity my reasoning behind that was not for a game wide mouse sensitivity or anything like that, more of a way to change the sensitivity the mouse is moved in first person in a script, for instance ads on a gun.

einsteinK:
Having a lot of built-in features, used (a lot) or not, isn't that bad of an idea.
Developers are encouraged to copy those scripts, improve and re-improve them.
With some code snippets on how to do certain stuff, editing will be a lot easier.

PR #119, opened 2015-02-20
Custom chat GUI support

samfun123:
**This pull request is a part of a feature request found [here](https://github.com/Anaminus/roblox-bug-tracker/issues/567), please read that before accepting this pull request!**

This would implement the EnumItems ChatWindow and ChatBar into the chat GUI. These need to be added to the API first which is explained on the feature request.


SolarCrane:
Pull requests are for fixes to bugs in the core scripts. If you want to suggest new APIs or features then please use the Roblox forums.

PR #122, opened 2015-02-20
Fixing studio issue

einsteinK:
"OnClientEvent can only be used on the client"


jmargh:
The script will never get to that code because of https://github.com/einsteinK/Core-Scripts/blob/patch-8/CoreScriptsRoot/CoreScripts/PlayerListScript2.lua#L20-L22. Maybe we can come up with a better solution to stop this scripts from loading/running while in studio.

einsteinK:
If I remember correctly, the person who had this error, had it also in Play Solo.

Maybe let the StarterScript load corescripts, depending on variables as being online, ...
The only issue would be for plugins (or cmdbar) using Players:CreateLocalPlayer(0)
(Which is a nice fast way of activating Play Solo while having normal studio features)

Issue #132, opened 2015-03-18
RequestFriendship should be passed a Player


TheSneak:
Source: `CoreGui.RobloxGui.CoreScripts/PlayerListScript2`

Most recently this error occurs on line 1047.

jmargh:
Verified. Will work on a fix for this.

PR #133, opened 2015-03-19
Fixing #132: Race condition with sending friend request and player leavi...

jmargh:
...ng.


jmargh:
Need to fix one thing and also update the top bar module.

PR #136, opened 2015-03-21
Fixing visual issue where items that can be purchased for both currencie...

jmargh:
...s,

but where the dev selected tix, the visual would show robux not tix.


jmargh:
Will be updating the pull request. Does not currently listen to the HeadsUpDisplay property on the vehicle seat being changed.

jmargh:
Reviewed with Ben.

PR #141, opened 2015-03-25
Added getVehicleSeat()

einsteinK:
Added a much better way of finding the VehicleSeat.
Works perfectly fine, unless a HumanoidRootPart isn't present.
(In that  case, it isn't a new (player) character, but an old one)

### DIFF
--- a/CoreScriptsRoot/CoreScripts/VehicleHud.lua
+++ b/CoreScriptsRoot/CoreScripts/VehicleHud.lua
@@ -103,12 +103,34 @@ local function onVehicleSeatChanged(property)
 	end
 end
 
+local function getVehicleSeat(part)
+	-- Same as GetConnectedParts(false)
+	-- Returns a list of all directly connected parts
+	-- (Those include parts directly welded to it)
+	-- Won't count other VSeats welded to the same vehicle
+	-- Could check for SeatWeld and its Part1, but
+	-- that isn't needed, this works fine already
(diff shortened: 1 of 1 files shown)


jmargh:
Sorry for taking a while to get to this. I didn't accept this one because we had plans to add new API for the humanoid to get the seat the humanoid is sitting in. The new API has been merged into our integration branch so I have changed this script to account for that. There will be a new SeatPart property on the humanoid and the .Seated event will now give you the seat the humanoid is sitting in as the second parameter, nil if getting out of a seat.

While both approaches (GetConnectedParts and cameraSubject being a VehicleSeat) work, we wanted to make it a bit more intuitive to find the seat. We also have plans to move the default vehicle controls to lua, so this will also help with that.

einsteinK:
That's better, of course.
The way I approached just seemed to be safer, not depending on the camera actually
changing its CameraSubject to the seat, on or not on time.

AlexwareRBLX:
Instead of adding TONS of crap inside the character,  which will only increase slow gameplay (as you notice,  a server of about 100 players gets anyqjere from 10-20 FPS on average,  with terrible ping.  However,  take all HumanoidRootParts out of them,  the ping is better tremendously and the average FPS is 20-40 FPS)  Which brings me to my next question,  why cant the roblox coding and physics department (with 5+ years of experience) cant hook a camera nor seat detection nor gravity to the torso?

PR #142, opened 2015-03-31
Changed size of mainBackgroundContainer

Usering:
Both width and height were set to a scale of 1, but since the ChatBar shows up, there are extra pixels not accounted for. Just modified it so it does cover the background of the chatBar.


jmargh:
A couple things here. The bottom chat bar is 20 pixels, so extending it by 50 will cut off part of the loading screen. This will also change when the top bar is released, as there will no longer be a chat bar on the bottom of the screen. The top bar will take the top 32 pixels.

jmargh:
This will be fixed from 5f482c35874e00be33393b3fbd7d568cc634c405. This change also correctly adjust size and position of mainBackgroundContainer for the top bar.

PR #155, opened 2015-04-29
Refactor ChatScript2

Quenty:
There are several small changes to the code to make it easier to read.

* Change `Player` (which references Player.LocalPlayer) --> `LocalPlayer`
* Align the administrator list and the service list for readability
* Move FocusNow variable recording into a smaller scope as it wasn't needed in the higher one.


Quenty:
Ok. Git works differently than I expected, and changes I made to the repository after submitting the pull request still are considered in the pull request. 

Here are the notes on the OOP closure-based to OOP metatable based code.
This update does a few things. It may seem like a lot of code, but it’s
really just refactoring variables.
* Removes the :Initialize() block, which is a bad OOP practice
* Starts to avoid mixing logic and instantiation.
     * Note several more changes will be made soon to prevent
construction inside of the window itself
     * This is considered good OOP design
 * Allows the extension of the ChatWindow through inheritance

Several more changes will be needed to entirely decouple the chat
window from globals, but this is a good start.

I understand if ROBLOX doesn’t actually want this change, but as far as
code quality goes, this is a good step forwards.

Part of the goal here is to make sure that the code written here isn't just usable for one thing (a ROBLOX chat window), but can be reused as a killfeed code or something else. Right now, this code is violating several OOP design principles.

SolarCrane:
Overall I like the changes in this commit and want to take some of them, but there are a few issues:
1. I do not believe in lining-up equal signs for these [reasons](http://stackoverflow.com/a/101984).
2. That chat script is not being edited anymore since we are switching to using modules. Here is the new [Chat Module Script](https://github.com/ROBLOX/Core-Scripts/blob/master/CoreScriptsRoot/Modules/Chat.lua).
3. We prefer to take Pull Requests as one commit rather than multiple different commits.

Quenty:
Thanks! I'll keep these notes in mind for when I submit another pull request. Is there a reason this legacy code is being kept in the repository? 

Also, I would love to check to make sure that the movement from closure-based OOP to metatable based OOP is ok. Some feedback on that would be great.

SolarCrane:
We are keeping the legacy code as a fallback in-case there is an issue with changes on production environment. The plan is to eventually delete the legacy code when the new code has stood the test of time.

As far as closure-based OOP to metatable based OOP is concerned I will have to look at both and evaluate them. They both seem like pretty good approaches, so I will have to think on it.

Thanks,
SolarCrane

PR #156, opened 2015-04-30
Removed trailing whitespace

Phalanxia:


JohnMHarrisJr:
Thank you. Just........Thank you.

Only20:
We do try to stick to a unified style but we do not have strict rules on whitespace.
As such, we will not accept requests to conform to a particular style on a whim.

I am closing this, but that does not mean we are closed to suggestions, even on style.
(E.g. see discussion in #155)

Phalanxia:
Trailing whitespace is basically file noise. Removing it doesn't change the functionality of the scripts at all. One reason for removing it is it takes up unnecessary storage space. Another reason, especially for interpreted languages (which Lua is), is that parsing of unnecessary whitespace may affect the speed of compilation and execution. I understand not accepting this commit, but please do reconsider.

Lots of modern editors actually automatically remove trailing whitespace, as its almost never intentional. Editors that don't have it built in usually have extensions available to do the job:
- Visual Studio: https://visualstudiogallery.msdn.microsoft.com/a204e29b-1778-4dae-affd-209bea658a59 
- Sublime Text: https://github.com/SublimeText/TrailingSpaces
- Emacs: http://emacswiki.org/emacs/DeletingWhitespace
- Atom: https://atom.io/packages/trailing-spaces

Related:
- https://programmers.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal
- http://www.dinduks.com/why-are-trailing-whitespaces-bad/

PR #157, opened 2015-05-03
Properly use the new API

einsteinK:
creatorFlagEnabled is always true due to the API being enabled.
This results in isCreator being true too, acting as if the player owns the current place.
Apparently, actually using the new API wasn't implemented here.

This commit should switch from checking the game's CreatorId to the new API, if it's (still) disabled.
Proper error handling is included. When the new method fails, it'll resort to the old method again.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/DeveloperConsole.lua
+++ b/CoreScriptsRoot/CoreScripts/DeveloperConsole.lua
@@ -1,6 +1,9 @@
 --Include
 local Create = assert(LoadLibrary("RbxUtility")).Create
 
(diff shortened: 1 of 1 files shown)


einsteinK:
Commandbar is still visible in all places for everyone.

einsteinK:
Also serverside internally, LogService:ExecuteScript(source) doesn't seem to work in group games.
I assume that it doesn't check if the player using the commandbar can edit the (group) place.

matthewdean:
Thanks for pointing this out. It will be fixed with the release of the new developer console.

PR #162, opened 2015-05-15
Fixing error

einsteinK:
Not sure if there was an issue open for this, since this error happened a lot due to several reasons.
This is another case where the BlackFrame error happens, due to the wrong checking for if a child is present.


TheSneak:
Heh, I submitted this same change here: https://github.com/ROBLOX/Core-Scripts/pull/146/files

einsteinK:
Seems like I'm blind

PR #165, opened 2015-06-01
Fixes keyboard input canceling movement when opposing keys are pressed

spacek531:
The current behavior of player scripts is that when opposing keys are held (such as W and S) the player stops moving. This is quite jarring and dissimilar to Roblox's controls before player scripts and dissimilar to most FPS games.

This update changes behavior by using leftValue, forwardValue, rightValue, and backwardValue as indicators that a key is pressed, and uses two different directional values (one for the X axis, and one for the Z axis) to determine walk direction. When a key is pressed, the direction value associated with the key gets forced into the key's direction. When the key is released, the code checks to see if the opposite key is being held to determine what the direction value should be.


OsyrisRBLX:
This is unexpected behavior. I don't think it should be implemented. For those who need it, they can implement it in their own games.

That's why we have PlayerScripts in the first place, isn't it?

spacek531:
@OsyrisRBLX it is *not* unexpected because, as I said in the description, *Roblox did this before.* The point of player scripts is to re-create what roblox did before, not change it.

Gskartwii:
This still is an issue.

PR #173, opened 2015-06-16
Zoom+pan fix

spacek531:
Hotfix for UserInputService bug where scrolling will revert UIS.MouseBehavior to Default. bfcf778 is the only intended commit.


spacek531:
The bug is fixed, so closing the pull request.

PR #191, opened 2015-07-18
Fix DevMovementMode Jump Bug

ConorGriffin37:
When DevComputerMovementMode is changed isJumping is not reset. isJumping should be set to false when a module is reset, like what is done with the movement vector.


ConorGriffin37:
This bug was reported here : 

http://developer.roblox.com/forum/client-bug-reports/16861-devmovementmode-jump-bug

hockeychaos:
can you recommit with just the jump bug fix?

PR #212, opened 2015-08-20
Ability to disable hurt overlay

DarraghGriffin:


DarraghGriffin:
Fixed in another way.

PR #231, opened 2015-09-10
Add "mute" functionality to the chat

DarraghGriffin:
Mute would allow players to stop receiving messages from a player but
only for that current game.


hockeychaos:
looks good, but two small things:
1) line 2410 in chat script should say unmute, not mute
2) can we change the data structure in MutedList to store the userId as the key, and the value just be a bool? Would make lookup code shorter/faster.

DarraghGriffin:
Made those changes and also changed the data structure used for blocking as well

PR #245, opened 2015-10-23
Remove Teleport Messages and Old Teleport Confirmation Dialog

DarraghGriffin:
Removes the Teleport Messages. Also removed the old teleport dialog, it is no longer used.


hockeychaos:
Ok so this makes sense, but it does seem like a big enough change to warrant a flag. Also, is there a reason to use coroutines instead of spawn? I think spawn is a bit better as you don't have to remember to call resume, since it looks like we always start the coroutine right after creating it.

DarraghGriffin:
Ok, cool, I'll add the flag tomorrow. 

I used coroutines because I wanted to pass an argument to the function for the coroutinesMap. I'm sure I could achieve the same thing a different way if that's preferable though. Coroutines also don't yield before executing but I don't think that difference from spawn is needed in this case.

PR #248, opened 2015-10-29
[Feature&Fix] Disable Reset button when needed

einsteinK:
If the developer disabled resetting by using Humanoid.SetStateEnabled, the reset button will be hidden.
The text above will also change its message to "The game doesn't allow you to reset!".

Small fix: Store LocalPlayer in a variable and prevent errors from calling Players-service something different than "Players"

### DIFF
--- a/CoreScriptsRoot/Modules/Settings/Pages/ResetCharacter.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/ResetCharacter.lua
@@ -15,7 +15,13 @@ local CoreGui = game:GetService("CoreGui")
 local ContextActionService = game:GetService("ContextActionService")
 local RobloxGui = CoreGui:WaitForChild("RobloxGui")
 local GuiService = game:GetService("GuiService")
-local Players = game:GetService("Players)
+local Players = game:GetService("Players")
+
+------------ Variables -------------------
+local PageInstance = nil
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/Settings/Pages/ResetCharacter.lua
```diff
@@ -7,23 +7,36 @@
 
 -------------- CONSTANTS -------------
 local RESET_CHARACTER_GAME_ACTION = "ResetCharacterAction"
+local RESET_ENABLED_TEXT = "Are you sure you want to reset your character?"
+local RESET_DISABLED_TEXT = "The game doesn't allow you to reset!"
```
DarraghGriffin: RESET_DISABLED_TEXT should probably be something else. canResetChar() returning false doesn't mean that the game doesn't allow you to reset, it only means you can't currently reset.

einsteinK: Yeah, something like "You can currently not reset!" is maybe better.
That's something that can always be changed when needed.

hockeychaos:
Instead of worrying about the messaging, how about just removing the reset button? I know we do this for stuff like mouse lock, but going forward I think we should just remove the option.

PR #250, opened 2015-11-18
Remember pevious GuiNavigationEnabled value in menu

DarraghGriffin:
More details on this bug : http://devforum.roblox.com/t/escape-menu-doenst-revert-changes-made-to-guiservice-on-close/19938


hockeychaos:
is there a reason the variable is not initialized to the current guiservice value?

hockeychaos:
we will also need to keep track of any dev changes to this variable, as a script could be changing this as we are in the menu.  Better yet, maybe we don't need to change this value at all in this script?

DarraghGriffin:
Can you check if we don't need to change the value at all in this script? I don't have a gamepad so I can't check if it has an impact on the menu.

PR #267, opened 2016-01-20
SetCore TopbarEnabled

Stickmasterluke:
Allows developers to call StarterGui:SetCore("TopbarEnabled",false) to turn off the topbar and all core UI except the setting hamburger icon.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/Topbar.lua
+++ b/CoreScriptsRoot/CoreScripts/Topbar.lua
@@ -36,6 +36,10 @@ local function GetChatVisibleIconFlag()
 	local chatVisibleIconSuccess, chatVisibleIconFlagValue = pcall(function() return settings():GetFFlag("MobileToggleChatVisibleIcon") end)
 	return chatVisibleIconSuccess and chatVisibleIconFlagValue == true
 end
+
+local defeatableTopbarSuccess, defeatableTopbarFlagValue = pcall(function() return settings():GetFFlag("EnableSetCoreTopbarEnabled") end)
+local defeatableTopbar = (defeatableTopbarSuccess and defeatableTopbarFlagValue == true)
+
 --[[ END OF FFLAG VALUES ]]
 
 
@@ -49,6 +53,11 @@ local StarterGui = game:GetService('StarterGui')
 
 --[[ END OF SERVICES ]]
 
+
+local topbarEnabled = true
+local topbarEnabledChangedEvent = Instance.new('BindableEvent')
+
+local settingsActive = false
 
 local GameSettings = UserSettings().GameSettings
 local Player = PlayersService.LocalPlayer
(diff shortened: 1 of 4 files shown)


CoreScriptsRoot/Modules/PlayerlistModule.lua
```diff
@@ -1703,8 +1708,14 @@ local function onCoreGuiChanged(coreGuiType, enabled)
 	end
 end
 
-onCoreGuiChanged(Enum.CoreGuiType.PlayerList, game:GetService("StarterGui"):GetCoreGuiEnabled(Enum.CoreGuiType.PlayerList))
-game:GetService("StarterGui").CoreGuiChangedSignal:connect(onCoreGuiChanged)
+function Playerlist:TopbarEnabledChanged(enabled)
```
SolarCrane: Since other functions in Playerlist are using the '.' instead of ':' syntax and because Playerlist is a singleton we should continue the convention and use a '.' here for the function declaration. Please revise other scripts to match this change.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -161,8 +166,8 @@ local function CreateTopBar()
 	end
 
 	function this:SetTopbarDisplayMode(opaque)
-		topbarContainer.BackgroundTransparency = opaque and TOPBAR_OPAQUE_TRANSPARENCY or TOPBAR_TRANSLUCENT_TRANSPARENCY
-		topbarShadow.Visible = not opaque
+		topbarContainer.BackgroundTransparency = (not topbarEnabled and 1) or (opaque and TOPBAR_OPAQUE_TRANSPARENCY) or TOPBAR_TRANSLUCENT_TRANSPARENCY
```
SolarCrane: This line and the above BackgroundTransparency lines are getting out of hand.
I suggest we create an extra function in Topbar called ComputeTransparency() to do all this calculations in an easy to read way.

bmanre2:
THIS IS THE HOLY SAVIOR OF ALL DEVS!

Gskartwii:
+1

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -36,6 +36,10 @@ local function GetChatVisibleIconFlag()
 	local chatVisibleIconSuccess, chatVisibleIconFlagValue = pcall(function() return settings():GetFFlag("MobileToggleChatVisibleIcon") end)
 	return chatVisibleIconSuccess and chatVisibleIconFlagValue == true
 end
+
+local disableableTopbarSuccess, disableableTopbarFlagValue = pcall(function() return settings():GetFFlag("EnableSetCoreTopbarEnabled") end)
+local disableableTopbar = (disableableTopbarSuccess and disableableTopbarFlagValue == true)
```
SolarCrane: I'm not sure if disableable is a word, can we rename this variable so that it is a little more understandable?

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -151,7 +159,7 @@ local function CreateTopBar()
 	function this:UpdateBackgroundTransparency()
 		local playerGui = Player:FindFirstChild('PlayerGui')
 		if playerGui then
-			topbarContainer.BackgroundTransparency = playerGui:GetTopbarTransparency()
+			topbarContainer.BackgroundTransparency = (not topbarEnabled and 1) or playerGui:GetTopbarTransparency()
```
SolarCrane: Now that we have ComputeTransparency lets move this logic into that function as well. This will make it less likely to run into some weird cases.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -160,9 +168,18 @@ local function CreateTopBar()
...
+		return TOPBAR_TRANSLUCENT_TRANSPARENCY
+	end
+
 	function this:SetTopbarDisplayMode(opaque)
-		topbarContainer.BackgroundTransparency = opaque and TOPBAR_OPAQUE_TRANSPARENCY or TOPBAR_TRANSLUCENT_TRANSPARENCY
-		topbarShadow.Visible = not opaque
+		topbarContainer.BackgroundTransparency = ComputeTransparency(opaque)
+		topbarShadow.Visible = topbarEnabled and not opaque
 		this:UpdateBackgroundTransparency()
 	end
```
SolarCrane: I missed this before, but since ForceTopbarOpaque is also called by the hamburger menu we probably need to incorporate this into our topbarEnabled code

bmanre2:
yASSS!!! IT HAS BEEN DONE!

Zero-Bits:
Great, so when will it be in-game?

Gskartwii:
@Zero-Bits As it was already merged, I would think right now, or at least soon enough.

Zero-Bits:
Looks like it's not available yet in studio.

Gskartwii:
I'm going to say that the "FFlag" is not set yet, but don't quote me on that.

PR #272, opened 2016-01-26
Ethan2

ethankoester:
Ethan2


CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -963,10 +963,6 @@ local function CreateSettingsHub()
 
 			local noOpFunc = function() end
 			ContextActionService:BindCoreAction("RbxSettingsHubStopCharacter", noOpFunc, false,
-												 Enum.PlayerActions.CharacterForward,
```
hockeychaos: Whats the reasoning behind removing these? I think these stop the character from moving when in the menu?

PR #282, opened 2016-02-08
Barebones VR support, take two

kyle-emmerich:
Fixed the issue with a line being commented out! This is barebones VR support, just the bare minimum to somewhat prevent nausea when using VR on ROBLOX. EDIT: Follow is good, but avoid Watch.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript.rbxmx
@@ -138,6 +138,8 @@ end
 local function Update()
 	if EnabledCamera then
 		EnabledCamera:Update()
+		EnabledCamera:ApplyCameraCFrame()
+		EnabledCamera:ApplyVROffset()
 	end
 	if EnabledOcclusion then
 		EnabledOcclusion:Update()
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
@@ -287,7 +287,7 @@ local function CreateCamera()
 	end
 
 	function this:ApplyVROffset()
-		if UserInputService.IsVREnabled and workspace.CurrentCamera then
+		if UserInputService.IsVREnabled and workspace.CurrentCamera and workspace.CurrentCamera.CameraType ~= Enum.CameraType.Scriptable then
 			workspace.CurrentCamera.CoordinateFrame = workspace.CurrentCamera.CoordinateFrame * UserInputService.UserHeadCFrame
 		end
 	end
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
@@ -793,7 +793,7 @@ local function CreateCamera()
 				wait()
 				humanoid = findPlayerHumanoid(player)
 			end
-			local function setLookBehindChatacter()
+			local function setLookBehindCharacter()
 				if humanoid and humanoid.Torso and player.Character == newCharacter then
 					local newDesiredLook = (humanoid.Torso.CFrame.lookVector - Vector3.new(0,0.23,0)).unit
 					local horizontalShift = findAngleBetweenXZVectors(newDesiredLook, this:GetCameraLook())
@@ -811,7 +811,7 @@ local function CreateCamera()
 				end
 			end
 			wait()
-			setLookBehindChatacter()
+			setLookBehindCharacter()
 		end
 
 		player.CharacterAdded:connect(function(character)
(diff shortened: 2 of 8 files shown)


kyle-emmerich:
I tested and made sure that the camera behaves as it did before changing internal state handling. Seems to work okay, should we move ahead?

kyle-emmerich:
Fixed up Follow camera type.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera/FollowCamera.rbxmx
```diff
@@ -138,7 +139,7 @@ local function CreateFollowCamera()
 						local lastVec = -(self.LastCameraTransform.p - subjectPosition)
 						local y = findAngleBetweenXZVectors(lastVec, self:GetCameraLook())
 						-- Check for NaNs
-						if IsFinite(y) and math.abs(y) > 0.0001 then
+						if IsFinite(y) and math.abs(y) > 0.0001 and not UserInputService.IsVREnabled then
```
hockeychaos: check should either be in the elseif statement (no reason to do all the math to just to bail)

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -72,6 +73,8 @@ end
 local function CreateCamera()
 	local this = {}
 
+	this.cframe = workspace.CurrentCamera and workspace.CurrentCamera.CoordinateFrame or CFrame.new()
```
hockeychaos: what do we gain by abstracting cframe? It seems like it just adds a layer of direction that could be broken, and I'm not sure if there are any tangible performance benefits.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera/FollowCamera.rbxmx
kyle-emmerich: True enough, will fix

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
kyle-emmerich: The reasoning is that previously, the camera scripts would just work off the existing state of the camera object. If your VR offset is modifying that, your state is now invalid unless you roll it back somehow, which is prone to floating point precision errors and isn't very friendly if you need to mix the transforms in a different way. It's definitely necessary if we want reasonable VR code.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera/FollowCamera.rbxmx
```diff
@@ -134,7 +135,7 @@ local function CreateFollowCamera()
 								self.RotateInput = self.RotateInput + Vector2.new(y * percent, 0)
 							end
 						end
-					elseif not (isInFirstPerson or userRecentlyPannedCamera) then
+					elseif not (isInFirstPerson or userRecentlyPannedCamera) and UserInputService.IsVREnabled then
```
hockeychaos: think this is missing a not here

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript.rbxmx
```diff
@@ -138,6 +138,8 @@ end
 local function Update()
 	if EnabledCamera then
 		EnabledCamera:Update()
+		EnabledCamera:ApplyCameraCFrame()
+		EnabledCamera:ApplyVROffset()
```
BinaryResolved: Is :ApplyVROffset() a Lua function or a ROBLOX made one?

kyle-emmerich: It's a Lua function, you can find it in RootCamera. It basically just applies the UserHeadCFrame on top of the current camera CFrame, but the plan was for individual cameras to modify it for more special handling.

I would love to see these camera modules become more object-oriented. Right now most of them just repeat the same code.

BinaryResolved: I am thinking about getting a VR headset. What kind are you using?

kyle-emmerich: The Oculus Rift DK2 works best so far. The CV1 probably works, but of course we can't really verify yet.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera/FollowCamera.rbxmx
kyle-emmerich: Good catch, I'll stop with the derp commits some time soon

PR #286, opened 2016-02-08
Removed Print()'s

BinaryResolved:
Having CoreGui scripts print something in the output can be very confusing, especially for beginner scripters.

### DIFF
--- a/CoreScriptsRoot/ServerCoreScripts/ServerSocialScript.lua
+++ b/CoreScriptsRoot/ServerCoreScripts/ServerSocialScript.lua
@@ -38,7 +38,6 @@ local function decodeJSON(json)
 		return HttpService:JSONDecode(json)
 	end)
 	if not success then
-		print("decodeJSON() failed because", result, "Input:", json)
 		return nil
 	end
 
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/ServerCoreScripts/ServerSocialScript.lua
```diff
@@ -38,7 +38,6 @@ local function decodeJSON(json)
 		return HttpService:JSONDecode(json)
 	end)
 	if not success then
-		print("decodeJSON() failed because", result, "Input:", json)
```
hockeychaos: I'd like to keep this because it doesn't happen often and when we get errors it's good to log them

BinaryResolved: I have only seen it in studio mode. Maybe we could use a :IsStudio() and return if true, and remove the other two. As it is confusing especially when using printing something similar when sending HTTP. Or would you just like to leave alone?

jmargh: Ya, we saw the issue in studio when we turned this on last week. ac3d1de06ecbe7111a31d12874d77738ca5fe3b8 fixes this by returning early if we're in studio, before any endpoints are hit.

hockeychaos: yeah so lets keep this one, but remove the others

PR #292, opened 2016-02-10
Easier Custom Health Gui

BinaryResolved:
Feedback is wonderful. Opinions is great. Improvements is best. So basically I used the same system as SolarCrane did for the easier custom chat script.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/HealthScript.lua
+++ b/CoreScriptsRoot/CoreScripts/HealthScript.lua
@@ -1,23 +1,24 @@
 --[[ 
 	This script controls the gui the player sees in regards to his or her health.
-	Can be turned with Game.StarterGui:SetCoreGuiEnabled(Enum.CoreGuiType.Health,false)
 	Copyright ROBLOX 2014. Written by Ben Tkacheff.
 --]]
 
 ---------------------------------------------------------------------
 -- Initialize/Variables
-while not game do
-	wait(1/60)
-end
-while not game:GetService("Players") do
-	wait(1/60)
-end
+---------------------------------------------------------------------
 
-local useCoreHealthBar = false
-local success = pcall(function() useCoreHealthBar = game:GetService("Players"):GetUseCoreScriptHealthBar() end)
-if not success or not useCoreHealthBar then
-	return
-end
+repeat wait() until game
+repeat wait() until game:GetService("Players").LocalPlayer
+
+local Player = game:GetService("Players").LocalPlayer
+
+-- NOTE: If you want to use this health script in your own game:
+-- 1) Copy the contents of this file into a module.
+-- 2) Create a localscript and parent it to StarterGui.
+-- 3) In the localscript require() the chat module you made in step 1.
+-- 4) Set the following variable to true.
+
+local useDefault = false
 
 local currentHumanoid = nil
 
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/CoreScripts/HealthScript.lua
```diff
@@ -262,55 +260,71 @@ function newPlayerCharacter()
 end
 
 function startGui()
-	characterAddedConnection = game:GetService("Players").LocalPlayer.CharacterAdded:connect(newPlayerCharacter)
+	characterAddedConnection = Players.LocalPlayer.CharacterAdded:connect(newPlayerCharacter)
```
hockeychaos: Why does this not use Player?

CoreScriptsRoot/CoreScripts/HealthScript.lua
```diff
@@ -262,55 +260,71 @@ function newPlayerCharacter()
...
 
-	local character = game:GetService("Players").LocalPlayer.Character
+	local character = Players.LocalPlayer.Character
```
hockeychaos: Why does this not use Player?

CoreScriptsRoot/CoreScripts/HealthScript.lua
```diff
@@ -262,55 +260,71 @@ function newPlayerCharacter()
...
+			HealthGui = Instance.new("ScreenGui")
+			HealthGui.Name = "HealthGui"
+			HealthGui.Parent = Player.PlayerGui
+			startGui()
+		end
+	end
+end
 
-game:GetService("StarterGui").CoreGuiChangedSignal:connect(function(coreGuiType,enabled)
+pcall(function()StarterGui.CoreGuiChangedSignal:connect(function(coreGuiType,enabled)
```
hockeychaos: why is this pcalled?

CoreScriptsRoot/CoreScripts/HealthScript.lua
```diff
@@ -262,55 +260,71 @@ function newPlayerCharacter()
...
 end)
+end)
 
-if game:GetService("StarterGui"):GetCoreGuiEnabled(Enum.CoreGuiType.Health) then
-	guiEnabled = true
-	startGui()
-end
+begin()
+
+return true
```
hockeychaos: why do we return true at the end of this?

CoreScriptsRoot/CoreScripts/HealthScript.lua
BinaryResolved: This is pcalled because it can only be used in CoreGui's. And if a player was to try to use it in there game, it would error and the script would not run.

CoreScriptsRoot/CoreScripts/HealthScript.lua
BinaryResolved: Seems I missed it, good eye.

CoreScriptsRoot/CoreScripts/HealthScript.lua
BinaryResolved: Same as other one.

CoreScriptsRoot/CoreScripts/HealthScript.lua
BinaryResolved: Module scripts require you to return a value other wise you get this.
`16:11:22.087 - Module code did not return exactly one value`
So I put a return true to stop that from erroring. Why do you have a better way to do it?

kyle-emmerich: Since when was HealthScript a module? Not being sarcastic, actually asking.

BinaryResolved: I thought all core scripts where modules. That had localscripts require() them. Am I wrong?

hockeychaos: Not all corescripts are modules, just the ones in the modules folder.

BinaryResolved: *facepalm*

BinaryResolved:
Sorry for the lack of finesse, I am still getting used to how this works.

PR #294, opened 2016-02-10
Attempt to fix issue: 163.

BinaryResolved:


CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1009,7 +1009,11 @@ function changeSlot(slot)
 			slot.Frame.BorderSizePixel = 3
 		end
 	else
-		slot:Select()
+		if slot:Select() ~= nil then --Attempt to fix a issue with Select being a nil value. Issue: 163
```
hockeychaos: This will end up trying to call slot:Select(), so you'll get the same error as before.

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1009,7 +1009,11 @@ function changeSlot(slot)
...
+			slot:Select()
+		else
+			return nil
```
hockeychaos: don't see why we should return nil here

PR #299, opened 2016-02-11
Attempt to fix issue: 293

BinaryResolved:
Attempt to fix issue: 293.

### DIFF
--- a/CoreScriptsRoot/Modules/Settings/Pages/Players.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/Players.lua
@@ -272,9 +272,11 @@ local function Initialize()
 					table.insert(existingPlayerLabels, index, frame)
 				end
 				frame.Name = 'PlayerLabel'..player.Name
-				frame.Icon.Image = 'http://www.roblox.com/Thumbs/Avatar.ashx?x=100&y=100&userId='..math.max(1, player.userId)
-				frame.NameLabel.Text = player.Name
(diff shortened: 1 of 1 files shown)


TheSneak:
This seems more appropriate to me:

    if frame:FindFirstChild("Icon") then
    	frame.Icon.Image = 'http://www.roblox.com/Thumbs/Avatar.ashx?x=100&y=100&userId='..math.max(1, player.userId)
    end
    if frame:FindFirstChild("NameLabel") then
    	frame.NameLabel.Text = player.Name
    end
    frame.ImageTransparency = frameDefaultTransparency

But a better fix would address the question "Why does Icon sometimes not exist as a child?"

BinaryResolved:
What if then 'NameLabel' is a valid member of frame but 'Icon' is not. Then it would show the name and not the Picture. Actually what if we did
`if frame:FindFirstChild("Icon") and frame:FindFirstChild("NameLabel") then
     frame.Icon.Image = 'http://www.roblox.com/Thumbs/Avatar.ashx?x=100&y=100&userId='..math.max(1, player.userId)
frame.NameLabel.Text = player.Name
`
What do you think of that?

hockeychaos:
I'm not sure if this will actually fix anything, so for now I'm going to close this.

BinaryResolved:
Well, the attempt was thrown detect if it was the child. If so then continue, if not then stop and will not error out the script.

BinaryResolved:
*to

PR #300, opened 2016-02-12
Make input at topbar work when hidden

einsteinK:
A quick and easy change that makes (regular PlayerGui) GuiObjects accept input.
They'll only receive input when the topbar is hidden using SetCore.
This allows people to create their own interactive menubar.

The settingsbutton still has Active set to true, so buttons under it don't accept input.
This change only has effect on stuff under the topbar where nothing is to begin with.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/Topbar.lua
+++ b/CoreScriptsRoot/CoreScripts/Topbar.lua
@@ -180,6 +180,10 @@ local function CreateTopBar()
 		end
 	end
 
+	function this:SetActive(active)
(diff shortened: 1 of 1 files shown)


hockeychaos:
This is on purpose for the time being

eLunate:
"_On purpose_" normally means "We didn't change it and we plan on keeping it this way for some reason we can't exactly tell you because _it's not valid_ "

einsteinK:
Not sure if I should delete this branch yet...

PR #302, opened 2016-02-12
Fix AlreadyOwn dialog

coolestkid:
Fixes the already owned dialog not closing when the OK button is pressed.


hockeychaos:
This looks like a risky fix, generally we are trying to keep any purchase fixes internal to ROBLOX. Could you give me repro steps as to what is breaking right now? Sounds like if you try to buy something you already own the dialog is not dismissable?

coolestkid:
Yes, if you attempt to purchase an asset you already own, the ok button won't close the dialog.

PR #304, opened 2016-02-12
Don't Buy Robux In Studio

coolestkid:
Don't Buy Robux In Studio to prevent hang.


hockeychaos:
what is the current behavior like when you try and buy robux in studio?

coolestkid:
It will show the purchase animation for the dialog, but since the browser window doesn't show in studio,  the animation will never stop.

PR #305, opened 2016-02-12
Fix race condition

coolestkid:
Fix race condition.


hockeychaos:
This change ends up clearing the purchase data right before we use it in the signal for marketplace service

coolestkid:
The call can be moved under the signal, and then this issue should be fixed.

PR #306, opened 2016-02-12
Fix error message

coolestkid:
Fix error message for cases when asset name was not retrieved.


hockeychaos:
What was broken about the way this worked? Is there a repro to show some bad behavior?

PR #310, opened 2016-02-15
Fix keyboard shortcuts

coolestkid:
Fix playerlist being incorrectly toggled when certain keyboard shortcuts are pressed.


hockeychaos:
This seems like a highly specialized fix that should be applied at all keyboard navigation thru guis at the engine level, I'll take a look.

PR #312, opened 2016-02-18
Hide Bubble Chat from Blocked Users

Stickmasterluke:


CoreScriptsRoot/CoreScripts/BubbleChat.lua
```diff
@@ -29,15 +29,17 @@ local BILLBOARD_MAX_HEIGHT = 500
 local ELIPSES = "..."
 local CchMaxChatMessageLength = 128 -- max chat message length, including null terminator and elipses.
 local CchMaxChatMessageLengthExclusive = CchMaxChatMessageLength - string.len(ELIPSES) - 1
+
+local NON_CORESCRIPT_MODE = false
```
hockeychaos: have you tested this script to see if it currently works in a non core script mode?

BinaryResolved: This would be very useful. Especially in big games with spammers, where you can still see Bubble Chat.

CoreScriptsRoot/CoreScripts/BubbleChat.lua
```diff
@@ -48,6 +50,13 @@ local BubbleColor = {	WHITE = "dub",
 --[[ END OF SCRIPT ENUMS ]]
 
 
+if not NON_CORESCRIPT_MODE then
+	while PlayersService.LocalPlayer == nil do PlayersService.ChildAdded:wait() end		--eww
```
hockeychaos: I think this code will work, but technically LocalPlayer is property, so it isn't necessary for a child to be added for this property to change.

CoreScriptsRoot/CoreScripts/BubbleChat.lua
```diff
@@ -48,6 +50,13 @@ local BubbleColor = {	WHITE = "dub",
...
+	local GuiRoot = CoreGuiService:WaitForChild('RobloxGui')
+	playerDropDownModule = require(GuiRoot.Modules:WaitForChild("PlayerDropDown"))
+	blockingUtility = playerDropDownModule:CreateBlockingUtility()
```
hockeychaos: this does not have any sort of local scope, should have this

CoreScriptsRoot/CoreScripts/BubbleChat.lua
```diff
@@ -552,12 +561,17 @@ local function createChatOutput()
 	function this:OnPlayerChatMessage(chatType, sourcePlayer, message, targetPlayer)
 		if not this:BubbleChatEnabled() then return end
 
-		-- eliminate display of emotes
-		if string.find(message, "/e ") == 1 or string.find(message, "/emote ") == 1 then return end
+		-- eliminate display of commands
+		if string.sub(message, 1, 1) == '/' then return end
```
hockeychaos: all commands use a forward slash?

PR #314, opened 2016-02-22
VR Camera Control Improvements

kyle-emmerich:
- Applying VR offset *after* occlusion check eliminates weird behavior when clipping camera into parts
- Changed IsVREnabled references to new VREnabled
- Changed pitch min/max to 30 degrees instead of 80, goes with following change
- Negated pitching of the camera after VR offset by a variable fraction, set to 50% right now.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
@@ -9,9 +9,9 @@
 local PlayersService = game:GetService('Players')
 local UserInputService = game:GetService('UserInputService')
 
---NoVR will be false if the platform does not have the VREnabled property, because indexing it will throw an error.
+--NoVRAPI will be false if the platform does not have the VREnabled property, because indexing it will throw an error.
 local success = pcall(function() local _ = UserInputService.VREnabled end)
-local NoVR = not success
+local NoVRAPI = not success
 
 local CameraScript = script.Parent
 local ShiftLockController = require(CameraScript:WaitForChild('ShiftLockController'))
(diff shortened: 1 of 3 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -9,8 +9,8 @@
 local PlayersService = game:GetService('Players')
 local UserInputService = game:GetService('UserInputService')
 
---NoVR will be false if the platform does not have the IsVREnabled property, because indexing it will throw an error.
-local success = pcall(function() local _ = UserInputService.IsVREnabled end)
+--NoVR will be false if the platform does not have the VREnabled property, because indexing it will throw an error.
+local success = pcall(function() local _ = UserInputService.VREnabled end)
 local NoVR = not success
```
SolarCrane: The naming of the 'NoVR' variable is very confusing because it doesn't mean VREnabled is false it just means that the API has not shipped yet. Something more verbose like 'VRApiExists' would make a lot more sense and avoid potential future bugs.

kyle-emmerich: I'm inclined to agree, it was a hasty fix to make sure things didn't explode. I'll change that.

SolarCrane:
We CANNOT accept commits from an `unknown` author. You need to set up your git username and email and then fix-up your commits before we can consider merging the pull request. If you aren't sure how to do it there is a guide on the git first-time setup page: [https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup)

SolarCrane:
When fixing up old commits that are still pending in pull requests squash them down using rebase and the 'fix-up' command

kyle-emmerich:
Not sure what happened with the author; SourceTree seems to have lost my configuration at some point. Will fix.

PR #317, opened 2016-02-24
Fix issue for nil AssetTypeId case in setBuyMoreRobuxDialog()

jmargh:
I ran into this issue while testing purchasing of Game Passes on ST1 for xbox. This will nil check the assetTypeId to string conversion so the script will no longer crash on site test. This issue only occurred when being prompted with an upsale of ROBUX.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
+++ b/CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
@@ -151,7 +151,8 @@ local ASSET_TO_STRING = {
 	[31] = "Right Leg";
 	[32] = "Package";
 	[33] = "YouTube Video";
-	[34] = "Game Pass";
+	-- NOTE: GamePass and Plugin AssetTypeIds are different on ST1, ST2 and ST3
+	[34] = "Game Pass";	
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
```diff
@@ -692,8 +693,8 @@ local function setBuyMoreRobuxDialog(playerBalance)
 	local neededRobux = PurchaseData.CurrencyAmount - playerBalanceInt
 	local productInfo = PurchaseData.ProductInfo
 
-	local descriptionText = "You need "..formatNumber(neededRobux).." more ROBUX to buy the "..productInfo["Name"].." "..
-		ASSET_TO_STRING[productInfo["AssetTypeId"]]
+	local descriptionText = "You need %s more ROBUX to buy the %s %s"
+	descriptionText = string.format(descriptionText, formatNumber(neededRobux), formatNumber(productInfo["Name"]), ASSET_TO_STRING[productInfo["AssetTypeId"]] or "")
```
hockeychaos: does formatNumber work for productInfo["Name"]? or at the least it seems a little weird

PR #319, opened 2016-02-27
Fixed possible memory leak in userList

Stickmasterluke:


CoreScriptsRoot/Modules/Settings/Pages/Players.lua
```diff
@@ -106,7 +106,6 @@ local function Initialize()
 			local friendLabelText = nil
 			if not status then
 				friendLabel = Instance.new('TextButton')
-				friendLabel.Name = 'FriendStatus'
```
hockeychaos: why did we remove the name?

CoreScriptsRoot/Modules/Settings/Pages/Players.lua
```diff
@@ -215,7 +214,7 @@ local function Initialize()
 
 	local existingPlayerLabels = {}
 	this.Displayed.Event:connect(function(switchedFromGamepadInput)
-		local sortedPlayers = PlayersService:GetPlayers()
```
hockeychaos: any reason to stop using PlayersService

CoreScriptsRoot/Modules/Settings/Pages/Players.lua
```diff
@@ -226,12 +225,15 @@ local function Initialize()
...
+
+
+		-- iterate through players to reuse or create labels for players
+		for index=1, #sortedPlayers do
 			local player = sortedPlayers[index]
 			local frame = existingPlayerLabels[index]
 			if player then
-				if not frame then
+				-- create label (frame) for this player index if one does not exist
+				if not frame or not frame.Parent then
```
hockeychaos: why do we create another frame just because the frame does not have a parent?

CoreScriptsRoot/Modules/Settings/Pages/Players.lua
```diff
@@ -215,7 +214,7 @@ local function Initialize()
 
 	local existingPlayerLabels = {}
```
hockeychaos: when is existingPlayerLabels actually populated?

CoreScriptsRoot/Modules/Settings/Pages/Players.lua
Stickmasterluke: Setting the name there is redundant, as the name is set later on in the code.

PR #324, opened 2016-03-02
VR Altitude of camera now dependent on a fixed angle, calculated usin…

kyle-emmerich:
…g zoom distance


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -155,11 +153,9 @@ local function CreateCamera()
 	end
 
 	function this:GetCameraHeight()
-		if this.currentHeight == nil then
-			this.currentHeight = clamp(MIN_HEIGHT, MAX_HEIGHT, 0)
-		end
-		if not this:IsInFirstPerson() then
```
hockeychaos: Looks like current height is no longer set here? do we no longer need it?

PR #326, opened 2016-03-04
Fixed problem with PopperCam & NPC CameraSubjects

CloneTrooper1019:
There was a bug with the PopperCam, where it would keep popping if the CameraSubject was set to an NPC, rather than the character of a Player.
Repro with the current version:
* Insert an NPC with a Humanoid into the Workspace.
* Set the CameraSubject to the NPC's Humanoid
* The PopperCam does not ignore the NPC, and will continuously keep popping forward.

This pull-request fixes the bug.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/PopperCam.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/PopperCam.rbxmx
@@ -153,7 +153,7 @@ function PopperCam:Update()
 		local ignoreList = {}
 		
 		local cameraSubject = Camera.CameraSubject
-		if cameraSubject:IsA("Humanoid") then
+		if cameraSubject and cameraSubject:IsA("Humanoid") then
 			local character = cameraSubject.Parent
 			if character and character:IsA("Model") then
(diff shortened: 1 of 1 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/PopperCam.rbxmx
```diff
@@ -151,6 +151,15 @@ function PopperCam:Update()
 		local cameraFrontPoint = cameraCFrame.p + (cameraCFrame.lookVector * NEAR_CLIP_PLANE_OFFSET)
 		local screenSize = Camera.ViewportSize
 		local ignoreList = {}
+		
+		local cameraSubject = Camera.CameraSubject
+		if cameraSubject:IsA("Humanoid") then
```
SolarCrane: What happens here if cameraSubject is nil? Does the script crash?

CloneTrooper1019: Ouch, good catch.
I'll fix that and resubmit.

hockeychaos:
closing because we need to change update to not construct ignore list every frame

PR #331, opened 2016-03-09
VR offset character with head position

kyle-emmerich:


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -294,6 +294,27 @@ local function CreateCamera()
...
+	function this:ApplyVRPosition()
+		--we only want this to happen in first person VR
+		local player = game.Players.LocalPlayer
+		if not player or not player.Character 
+				or not player.Character:FindFirstChild("HumanoidRootPart") 
+				or not player.Character.HumanoidRootPart:FindFirstChild("RootJoint") then
+			return
+		end
+		
+		if not UserInputService.VREnabled or not this:IsInFirstPerson() then
```
hockeychaos: It seems weird we are doing cframe adjustments in a function called ApplyVRPosition when VR is not enabled

kyle-emmerich: Would it be better if I moved the check for VR out of that function? I'm not sure how long this is going to stick around, but I agree that it should be as low-impact to non-VR as possible.

hockeychaos: why do we need to do this adjustment when you don't have vr?

PR #333, opened 2016-03-10
Remove flag + fix already purchased item

hockeychaos:
Removing UseNewPromptEndHandling because its been on for awhile.  Also
fixing an issue where if you are prompted to buy an asset you already
own, the prompt is not dismissable.


CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
```diff
@@ -1046,6 +1041,7 @@ local function canPurchase(disableUpsell)
 				onPurchaseFailed(PURCHASE_FAILED.DEFAULT_ERROR)
 				return false
 			end
+			purchaseState = PURCHASE_STATE.FAILED
```
SolarCrane: Maybe move this setting of `purchaseState` to `onPurchaseFailed` sometime?

PR #334, opened 2016-03-11
Fix Chat Commands

coolestkid:
Fix chat commands that have an underscore in them.


hockeychaos:
Merged a different pull request that also fixes this issue.

PR #335, opened 2016-03-11
Fix regex for usernames with underscores

DarraghGriffin:
With the current regex you can't use the chat commands /whisper, /block,
/mute ect on Players who have an underscore in their name. The %w
specifier is for alphanumberic characters only.

The regex (%w+_?%w+) is valid as usernames can have up to one underscore
and it can not be the first or last character in their name


hockeychaos:
Merging should be done by someone in the core script admin group FYI. This
merge seems fine, but in the future if you see a merge you feel good about
just let one of us know and we'll look at it.

On Fri, Mar 11, 2016 at 10:01 AM, ethankoester <notifications@github.com>
wrote:

> Merged #335 <https://github.com/ROBLOX/Core-Scripts/pull/335>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ROBLOX/Core-Scripts/pull/335#event-586747814>.
>



-- 
ben

PR #342, opened 2016-03-19
VR-23 Fixing bug where in VR your character would not rotate with you

SolarCrane:
looking direction when in first person


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -294,25 +294,25 @@ local function CreateCamera()
 
 	function this:Update()
```
BinaryResolved: Why have a empty function?

kyle-emmerich: If a camera type doesn't define it itself, it will use this. This ensures that it can be called without checking if it's nil. It's sort of an inheritance mechanism.

PR #353, opened 2016-03-26
VR-69 Healthbar in VR

kyle-emmerich:


CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -137,6 +162,7 @@ end
 
 local function OnCharacterAdded(character)
 	local backpack = player:WaitForChild("Backpack")
+	local humanoid = character:WaitForChild("Humanoid")
```
SolarCrane: Some characters in games never get a humanoid, which would make this callback yield forever.

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -153,6 +179,8 @@ local function OnCharacterAdded(character)
 	
 	backpack.ChildAdded:connect(OnChildAdded)
 	backpack.ChildRemoved:connect(OnChildRemoved)
+
+	humanoid.Changed:connect(function() UpdateHealth(humanoid) end)
```
SolarCrane: We should call update health here too, you might be given a half damaged character.

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: True. How do we handle this?

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: Fixed

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -39,29 +52,41 @@ local function UpdateLayout()
...
+	healthbarBack.Position = UDim2.new(0.5, -HEALTHBAR_WIDTH / 2, 0, (HEALTHBAR_SPACE - HEALTHBAR_HEIGHT) / 2)
+	healthbarBack.Size = UDim2.new(0, HEALTHBAR_WIDTH, 0, HEALTHBAR_HEIGHT)
 	
 	verticalRange = math.atan(hopperbinPart.Size.Y / (2 * HOPPERBIN_OFFSET.Z)) + math.rad(30) * 2
 	horizontalRange = math.atan(hopperbinPart.Size.X / (2 * HOPPERBIN_OFFSET.Z)) - math.rad(20) * 2
 end
 
+local function UpdateHealth(humanoid)
+	local percent = humanoid.Health / humanoid.MaxHealth
+	healthbarFront.Size = UDim2.new(percent, 0, 1, 0)
```
SolarCrane: There is also a color shift that we do in the regular health script base on your health, is it possible to do that here as well? Or is it not a desired feature?

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -29,6 +33,15 @@ hopperbinGUI.Adornee = hopperbinPart
...
 hopperbinGUI.Name = "HopperBinGUI"
 
+local healthbarBack = Instance.new("Frame", hopperbinGUI)
+healthbarBack.BackgroundColor3 = Color3.new(1, 0, 0)
+healthbarBack.BorderSizePixel = 0
+local healthbarFront = Instance.new("Frame", healthbarBack)
+healthbarFront.BorderSizePixel = 0
+healthbarFront.Size = UDim2.new(1, 0, 1, 0)
+healthbarFront.Position = UDim2.new(0, 0, 0, 0)
+healthbarFront.BackgroundColor3 = Color3.new(0, 1, 0)
```
SolarCrane: We should set the name property on these variables.

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -153,6 +179,11 @@ local function OnCharacterAdded(character)
 	
 	backpack.ChildAdded:connect(OnChildAdded)
 	backpack.ChildRemoved:connect(OnChildRemoved)
+
+	if humanoid then
+		humanoid.Changed:connect(function() UpdateHealth(humanoid) end)
+		UpdateHealth(humanoid)
+	end
```
SolarCrane: To be truly bulletproof we should account for the potential addition or removal of Humanoid from the character.

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -29,6 +33,15 @@ hopperbinGUI.Adornee = hopperbinPart
 hopperbinGUI.ToolPunchThroughDistance = 1000
 hopperbinGUI.Name = "HopperBinGUI"
 
+local healthbarBack = Instance.new("Frame", hopperbinGUI)
```
SolarCrane: This needs to show or hide based on the OnCoreGuiChanged for Healthbar coregui.
Also if backpack is not turned on but the Healthbar is turned on, does this module get required?

kyle-emmerich: I believe so, but I'll test it to make sure. The only check is for VR when requiring this, and all signs point to the code running no matter what's enabled.

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: Would you still believe me if I told you I've been making games on ROBLOX since 2007?

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: That is a good question. I would assume yes, we do want that, but we might touch bases with Keith and make sure.

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: Fixed

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: Fixed

CoreScriptsRoot/Modules/BackpackScript3D.lua
SolarCrane: Alright, this is something to circle back with design (they might have some idea for what it looks like)

kyle-emmerich: FWIW, I imitated the current design that's in the right side of the topbar for now.

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -28,40 +32,80 @@ local hopperbinGUI = Instance.new("SurfaceGui", CoreGui)
...
+local healthbarFront = Instance.new("Frame", healthbarBack)
+healthbarFront.BorderSizePixel = 0
+healthbarFront.Size = UDim2.new(1, 0, 1, 0)
+healthbarFront.Position = UDim2.new(0, 0, 0, 0)
+healthbarFront.BackgroundColor3 = Color3.new(0, 1, 0)
+healthbarFront.Name = "HealthbarFront"
+
+local greenColor = Color3.new(0.2, 1, 0.2)
+local redColor = Color3.new(1, 0.2, 0.2)
+local yellowColor = Color3.new(1, 1, 0.2)
```
SolarCrane: Since these variables are constants we should keep them consistent with our naming style such as: `GREEN_COLOR`

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -104,10 +148,25 @@ local function AddTool(tool)
...
+		humanoidChangedEvent = humanoid.Changed:connect(function() UpdateHealth(humanoid) end)
+		UpdateHealth(humanoid)
+	end
+end
+
 local function OnChildAdded(child)
 	if child:IsA("Tool") or child:IsA("HopperBin") then
 		AddTool(child)
 	end
+	if child:IsA("Humanoid") then
```
SolarCrane: Since `OnChildAdded` is used for backpack stuff too we might want to make sure child is in the character

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -104,10 +148,25 @@ local function AddTool(tool)
...
 end
 
+local humanoidChangedEvent = nil
+local function RegisterHumanoid(humanoid)
+	if humanoidChangedEvent then
+		humanoidChangedEvent:disconnect()
+		humanoidChangedEvent = nil
+	end
+	if humanoid then
+		humanoidChangedEvent = humanoid.Changed:connect(function() UpdateHealth(humanoid) end)
```
SolarCrane: What happens if this humanoid is de-parented from the character?

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: Hm, I copied & pasted them from HealthScript.lua.

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: Can't hurt, though I've never seen someone put a humanoid in their backpack...

CoreScriptsRoot/Modules/BackpackScript3D.lua
SolarCrane: There is also a special event for just health changes events on humanoid: `humanoid.HealthChanged`
This way we don't have to do a bunch of updates for when other properties change

kyle-emmerich: I could watch for Parent changing.

What happens to the healthbar if you remove the humanoid though? The way it currently functions is that it will just switch to whatever new humanoid you add after. Should the healthbar just disappear if there's NO humanoid?

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -137,6 +196,10 @@ end
 
 local function OnCharacterAdded(character)
 	local backpack = player:WaitForChild("Backpack")
+	local humanoid = character:FindFirstChild("Humanoid")
```
SolarCrane: This should be FindFirstChild by type not by name, correct?

kyle-emmerich: Correct, changed.

kyle-emmerich:
This is ready!

PR #354, opened 2016-03-28
Make leave button change color on mouse events

hockeychaos:
Currently the leave button doesn't look like it works on touch, because
the color doesn't change when you press the button, and it might take a
long time (up to 10 seconds) to actually close the game.  This is not a
great user experience, so this helps a bit by letting you know you
actually clicked the button


hockeychaos:
need global solution

PR #356, opened 2016-03-30
Bubblechat Nil Checks

Stickmasterluke:


CoreScriptsRoot/CoreScripts/BubbleChat.lua
```diff
@@ -334,7 +334,7 @@ local function createChatOutput()
 	end
 
 	function this:CreateBillboardGuiHelper(instance, onlyCharacter)
-		if not this.CharacterSortedMsg:Get(instance)["BillboardGui"] then
+		if instance and not this.CharacterSortedMsg:Get(instance)["BillboardGui"] then
```
hockeychaos: what happens if we just return if instance is nil?

CoreScriptsRoot/CoreScripts/BubbleChat.lua
```diff
@@ -133,7 +133,7 @@ local function createMap()
 	end
 
 	function this:Get(key)
-		if not this.data[key] then
+		if key and not this.data[key] then
```
hockeychaos: if key is nil, can we just return nil?

CoreScriptsRoot/CoreScripts/BubbleChat.lua
hockeychaos: bump

Stickmasterluke: No billboard is made or updated

PR #359, opened 2016-03-31
VR-72 VR Loading 3D Screen

SolarCrane:


CoreScriptsRoot/LoadingScript.lua
```diff
@@ -762,3 +763,24 @@ game:GetService("ReplicatedFirst").RemoveDefaultLoadingGuiSignal:connect(handleR
 if game:GetService("ReplicatedFirst"):IsDefaultLoadingGuiRemoved() then
 	handleRemoveDefaultLoadingGui()
 end
+
+local UserInputServiceChangedConn;
```
BinaryResolved: Why do we include a ";" at the end of it? I rarely see that after declaring the connection variable.

kyle-emmerich: Probably to denote that it's not being initialized (left nil)

SolarCrane: The reason to use a `;` at the end of the line is to prevent mistakes such as this:

local something


= "blagh"

In the above example something would be accidentally initialized to "blagh"

BinaryResolved: Oh, ok. Thanks for both of your replies. Much appreciated.

CoreScriptsRoot/LoadingScript.lua
```diff
@@ -762,3 +763,24 @@ game:GetService("ReplicatedFirst").RemoveDefaultLoadingGuiSignal:connect(handleR
...
+local function onUserInputServiceChanged(prop)
+	if prop == 'VREnabled' then
+		local UseVr = false
+		pcall(function() UseVr = UIS.VREnabled end)
+
+		if UseVr then
+			require(RobloxGui.Modules.LoadingScreen3D)
```
hockeychaos: is there any sort of 'pop' with replacing the loading screen if vrenabled changes? is this even possible right now?

CoreScriptsRoot/LoadingScript.lua
```diff
@@ -762,3 +763,24 @@ game:GetService("ReplicatedFirst").RemoveDefaultLoadingGuiSignal:connect(handleR
...
+			handleRemoveDefaultLoadingGui(true)
+			if UserInputServiceChangedConn then
```
hockeychaos: this seems like it would be better at the beginning of the if (its possible require could cause a yield). It shouldn't, but it could

CoreScriptsRoot/Modules/LoadingScreen3D.lua
```diff
@@ -0,0 +1,355 @@
...
+
+GameInfoProvider:LoadAssetsAsync()
+if GameInfoProvider:IsReady() then
+	OnGameInfoLoaded()
+end
+-- Disco the following events?
+GameInfoProvider.LoadingFinishedEvent:connect(OnGameInfoLoaded)
+
+
+ReplicatedFirst.FinishedReplicating:connect(OnReplicatingFinished)
```
hockeychaos: any reason not to check the IsFinishedReplicating first? We can not connect to finishedreplicating in this situation

CoreScriptsRoot/Modules/LoadingScreen3D.lua
```diff
@@ -0,0 +1,355 @@
...
+if ReplicatedFirst:IsFinishedReplicating() then
+	OnReplicatingFinished()
+end
+
+ReplicatedFirst.RemoveDefaultLoadingGuiSignal:connect(OnDefaultLoadingGuiRemoved)
```
hockeychaos: same as above

CoreScriptsRoot/Modules/LoadingScreen3D.lua
```diff
@@ -0,0 +1,355 @@
...
+local MarketPlaceService = game:GetService('MarketplaceService')
+local UserInputService = game:GetService('UserInputService')
+local ReplicatedFirst = game:GetService('ReplicatedFirst')
+local GuiService = game:GetService('GuiService')
+
+local RobloxGui = CoreGui:WaitForChild("RobloxGui")
+local Util = require(RobloxGui.Modules.Settings.Utility)
+
+
+local function FadeElements(element, newValue, tweeners, duration)
```
hockeychaos: whats the point of the tweeners? seems we always pass an empty table that has no reference

CoreScriptsRoot/Modules/LoadingScreen3D.lua
```diff
@@ -0,0 +1,355 @@
...
+}
+
+local loadingSurfaceGui = Util:Create'SurfaceGui'
+{
+	Name = "LoadingSurfaceGui";
+	Adornee = surfaceGuiAdorn;
+	ToolPunchThroughDistance = 1000;
+	CanvasSize = Vector2.new(500, 500);
+	Archivable = false;
+	-- Parent = surfaceGuiAdorn;
```
hockeychaos: remove plz

CoreScriptsRoot/Modules/LoadingScreen3D.lua
```diff
@@ -0,0 +1,355 @@
...
+					freeze = false
+				end)
+			else
+				creatorTextPosition = newX
+			end
+		end
+		creatorText.Position = UDim2.new(0, creatorTextPosition, 0, 0)
+	end
+
+	-- if gameNameText.TextFits then
```
hockeychaos: do we need this?

CoreScriptsRoot/Modules/LoadingScreen3D.lua
```diff
@@ -0,0 +1,355 @@
...
+		creatorText.Size = UDim2.new(1, 0, 1, 0)
+	elseif delta ~= nil then
+		creatorText.Size = UDim2.new(0, creatorText.TextBounds.X, 1, 0)
+		if not freeze then
+			local newX = (creatorTextPosition - delta * TEXT_SCROLL_SPEED)
+			if newX + creatorText.AbsoluteSize.X < creatorTextContainer.AbsoluteSize.X then
+				freeze = true
+				spawn(function()
+					Util:TweenProperty(creatorText, 'TextTransparency', creatorText.TextTransparency, 1, 1, Util:GetEaseInOutQuad())
+					wait(1.5)
```
hockeychaos: why are we waiting in render stepped?

kyle-emmerich: I'm surprised you caught that it was in RenderStepped but not that it was within a spawn()'d function

hockeychaos: not enough coffee apparently

On Fri, Apr 1, 2016 at 3:12 PM, Kyle Emmerich <notifications@github.com>
wrote:

> In CoreScriptsRoot/Modules/LoadingScreen3D.lua
> <https://github.com/ROBLOX/Core-Scripts/pull/359#discussion_r58275048>:
>
> > +		surfaceGuiAdorn.Parent = camera
> > +	end
> > +
> > +
> > +	if creatorText.TextBounds.X < creatorTextContainer.AbsoluteSize.X then
> > +		creatorText.Position = UDim2.new(0, 0, 0, 0)
> > +		creatorText.Size = UDim2.new(1, 0, 1, 0)
> > +	elseif delta ~= nil then
> > +		creatorText.Size = UDim2.new(0, creatorText.TextBounds.X, 1, 0)
> > +		if not freeze then
> > +			local newX = (creatorTextPosition - delta * TEXT_SCROLL_SPEED)
> > +			if newX + creatorText.AbsoluteSize.X < creatorTextContainer.AbsoluteSize.X then
> > +				freeze = true
> > +				spawn(function()
> > +					Util:TweenProperty(creatorText, 'TextTransparency', creatorText.TextTransparency, 1, 1, Util:GetEaseInOutQuad())
> > +					wait(1.5)
>
> I'm surprised you caught that it was in RenderStepped but not that it was
> within a spawn()'d function
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/ROBLOX/Core-Scripts/pull/359/files/c4e35034507b4aba6fa04a83bcfc45892edf1617#r58275048>
>



-- 
ben

hockeychaos: how many spawned functions are we going to make per second? essentially, how often is newX + creatorText.AbsoluteSize.X < creatorTextContainer.AbsoluteSize.X?

kyle-emmerich: More coffee! I see it setting freeze=true, so it checks that beforehand. So you won't get it more than every 1.5 seconds.

SolarCrane: @hockeychaos has a good point; it is not very clear code. I will see if I can find a more elegant way to write it.

CoreScriptsRoot/LoadingScript.lua
SolarCrane: There is some future-proofing at work here. There was some worry that the script would run before this property was set.

PR #360, opened 2016-04-03
Update Chat.lua - Improper whitespace filtering

Scarfacial:
Update to `Util.FilterUnprintableCharacters` function - Changing whitespace filter from defective match to all whitespace characters.

This will change all whitespace characters into a single space, preventing both whitespace spam (multiple spaces) and invalid whitespace characters (such as newlines and tabs)

For clarity, also renaming the local variable `PRINTABLE_CHARS` to a more fitting `UNPRINTABLE_CHARS` as the assigned string pattern matches all characters *besides* the printable ones.

### DIFF
--- a/CoreScriptsRoot/Modules/Chat.lua
+++ b/CoreScriptsRoot/Modules/Chat.lua
@@ -341,15 +341,15 @@ do
 		return TextSizeCache[text][font][sizeBounds][fontSize]
 	end
 
-	local PRINTABLE_CHARS = '[^' .. string.char(32) .. '-' ..  string.char(126) .. ']'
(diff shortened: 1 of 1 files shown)


SolarCrane:
We want to allow multiple whitespace characters in succession because there are some cases where you may want to denote a larger gap when chatting. We decided on limiting it to 7 whitespace characters at a point because too many spaces can be abused.

Scarfacial:
Ah alright, that makes sense then.

I'm not sure if you'd prefer I close this myself or not, so I'll just leave it up for the time being.

SolarCrane:
I will close it, we also plan on revisiting this section of the code very soon, so things may be changing anways.
As always, thanks for effort!

PR #362, opened 2016-04-05
VR-71 Move settings to 3D panel in VR; VR-66 and VR-99 for A button t…

kyle-emmerich:
…o activate VR GUIs, VR-68 move hamburger, VR-45 hide mouse if inactive

### DIFF
--- a/CoreScriptsRoot/CoreScripts/Topbar.lua
+++ b/CoreScriptsRoot/CoreScripts/Topbar.lua
@@ -718,12 +718,31 @@ local function CreateSettingsIcon(topBarInstance)
 	rawset(menuItem, "SetTransparency", function(self, transparency)
 		settingsIconImage.ImageTransparency = transparency
 	end)
+	rawset(menuItem, "SetImage", function(self, image)
+		settingsIconImage.Image = image
+	end)
+	rawset(menuItem, "SetSettingsActive", function(self, active)
+		settingsActive = active
+		MenuModule:ToggleVisibility(settingsActive)
+		UpdateHamburgerIcon()
+
+		return settingsActive
+	end)
 
 	return menuItem
 end
 
 local function Create3DSettingsIcon(topBarInstance)
 	local menuItem = CreateSettingsIcon(topBarInstance)
+
+	rawset(menuItem, "Hover", function(self, hovering)
+		if hovering then
+			self:SetImage("rbxasset://textures/ui/Menu/HamburgerDown.png")
+		else
+			self:SetImage("rbxasset://textures/ui/Menu/Hamburger.png")
+		end
+	end)
+
 	return menuItem
 end
 ------------
@@ -1226,14 +1245,42 @@ if nameAndHealthMenuItem and topbarEnabled and not isTenFootInterface then
 end
 
 local Panel3D = require(GuiRoot.Modules.Panel3D)
+
 local function MoveHamburgerTo3D()
 	LeftMenubar:RemoveItem(settingsIcon)
 
 	local settingsIcon3D = Create3DSettingsIcon(TopBar)
 
+	local function OnHamburger3DInput(actionName, state, inputObj)
+		if state ~= Enum.UserInputState.Begin then
+			return
+		end
+		settingsIcon3D:SetSettingsActive(true) --this button is only ever shown if the settings menu isn't already open, so it can only be true.
+	end
+
+	local eaterAction = game:GetService("HttpService"):GenerateGUID()
+	local ContextActionService = game:GetService("ContextActionService")
+	local function EnableHamburger3DInput(enable)
+		if enable then
+			ContextActionService:BindCoreAction("Hamburger3DInput", OnHamburger3DInput, false, Enum.KeyCode.Space, Enum.KeyCode.ButtonA)
+			ContextActionService:BindAction(eaterAction, function() end, false, Enum.KeyCode.Space, Enum.KeyCode.ButtonA)
+		else
+			ContextActionService:UnbindCoreAction("Hamburger3DInput")
+			ContextActionService:UnbindAction(eaterAction)
+		end
+	end
+
 	local panel = Panel3D.Get(Panel3D.Panels.Hamburger)
 	panel:ResizePixels(50, TOPBAR_THICKNESS)
 	panel:AddTransparencyCallback(function(transparency) settingsIcon3D:SetTransparency(transparency) end)
+	panel.OnMouseEnter = function() 
+		EnableHamburger3DInput(true) 
+		settingsIcon3D:Hover(true)
+	end
+	panel.OnMouseLeave = function() 
+		EnableHamburger3DInput(false) 
+		settingsIcon3D:Hover(false)
+	end
 	
 	settingsIcon3D.Parent = panel.gui
 end
--- a/CoreScriptsRoot/Modules/Panel3D.lua
+++ b/CoreScriptsRoot/Modules/Panel3D.lua
@@ -1,6 +1,10 @@
 --Panel3D: 3D GUI panels for VR
 --written by 0xBAADF00D
 local PIXELS_PER_STUD = 64
+local SETTINGS_DISTANCE = 3.5
+
+local CURSOR_HIDE_TIME = 2
+local CURSOR_FADE_TIME = 0.125
 
 local CoreGui = game:GetService('CoreGui')
 local RobloxGui = CoreGui:WaitForChild("RobloxGui")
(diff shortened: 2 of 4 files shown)


CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -85,13 +88,20 @@ function Panel3D.Get(panelId)
 
 		panel.transparencyCallbacks = {}
 
+		panel.OnMouseEnter = false
+		panel.OnMouseLeave = false
```
SolarCrane: Why are we settings these to false and not nil? I am getting nervous about the implications.

kyle-emmerich: We don't do that here? It's functionally similar to nil, but the key will be actually set on the table. Setting it to nil does nothing. Both methods will denote in the code that something is there, so the only difference is in using pairs().

SolarCrane: We usually set it to nil if you really want to be explicit or nothing at all and you get to save 2 lines of code!

BinaryResolved: Since they are the same value why don't we do something like.
`panel.OnMouseEnter, panel.OnMouseLeave = false`
They will still act like to different variables, but since they have the starting value it can save lines of code like SolarCrane said. Thoughts?

kyle-emmerich: It's probably not worth it at this point, and saving lines of code isn't a super high priority after all!

BinaryResolved: Hehe

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -9,36 +13,109 @@ local Panel3D = {}
...
+					return
+				end
+			end
+		end
+		cursorHidden = true
+		spawn(function()
+			pcall(function() 
+				while UserInputService.MouseIconTransparency < 1 and cursorHidden do
+					local dt = game:GetService("RunService").RenderStepped:wait()
+					UserInputService.MouseIconTransparency = UserInputService.MouseIconTransparency + dt / CURSOR_FADE_TIME
```
SolarCrane: I don't know what happens if you set `UserInputService.MouseIconTransparency` to greater than one but we should wrap this in a `math.min(1, yourcode)` block

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -9,36 +13,109 @@ local Panel3D = {}
...
+	end
+end)
+game:GetService("RunService").Heartbeat:connect(function()
+	if (UserInputService.MouseIconEnabled and UserInputService.OverrideMouseIconBehavior == Enum.OverrideMouseIconBehavior.None) or
+		UserInputService.OverrideMouseIconBehavior == Enum.OverrideMouseIconBehavior.ForceShow then
+		cursorHidden = false
+	end
+	pcall(function()
+		if UserInputService.MouseIconTransparency == 0 then
+			cursorHidden = false
```
SolarCrane: Why do we need this? is this not handle in the autoHideCursor function?

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -78,20 +156,34 @@ function Panel3D.Get(panelId)
 		panel.gui.Name = ("%sPanelGUI"):format(panelName)
 
 		panel.vector = panelDefaultVectors[panelId]
+		panel.overrideCFrame = nil
```
SolarCrane: I thought we were getting rid of overrideCFrame

CoreScriptsRoot/Modules/Panel3D.lua
kyle-emmerich: autoHideCursor won't be called unless cursorHidden = false. Sometimes other things modify OverrideMouseIconBehavior, MouseIconEnabled, or MouseIconTransparency so the mouse will be shown, but cursorHidden remains true.

Can you think of a better way to handle this?

CoreScriptsRoot/Modules/Panel3D.lua
kyle-emmerich: I don't think anything bad will happen, but sure.

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -9,36 +13,109 @@ local Panel3D = {}
...
+	if not UserInputService.VREnabled then
+		cursorHidden = false
+		UserInputService.MouseIconTransparency = 0
+		return
+	end
+	if hide then
+		--don't hide if there's a tool in the character
+		local character = game.Players.LocalPlayer.Character
+		if character then
+			for i, v in ipairs(character:GetChildren()) do
```
SolarCrane: This is a lot of code to do on render step! We can keep track of whether or not you have a tool equipped outside of this function.

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -9,36 +13,109 @@ local Panel3D = {}
...
+				if v:IsA("Tool") then
+					autoHideCursor(false)
+					lastMouseMove = tick()
```
SolarCrane: Is this really a mouseMove event? It seems like we are having a misleading variable name here.

CoreScriptsRoot/Modules/Panel3D.lua
kyle-emmerich: This should only happen once until cursorHidden=false and the timeout has occurred again.

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -9,36 +13,124 @@ local Panel3D = {}
...
+					return
+				end
+			end
+		end
+		cursorHidden = true
+		spawn(function()
```
SolarCrane: A cleaner way to do this would be to BindToRenderStep rather than a spawn

CoreScriptsRoot/Modules/Panel3D.lua
kyle-emmerich: Yes, it's updated on line 120 in an InputChanged event when UserInputType==MouseMovement

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -9,36 +13,124 @@ local Panel3D = {}
...
+					local dt = game:GetService("RunService").RenderStepped:wait()
+					UserInputService.MouseIconTransparency = math.min(1, UserInputService.MouseIconTransparency + dt / CURSOR_FADE_TIME)
+				end
+			end)
+			UserInputService.OverrideMouseIconBehavior = Enum.OverrideMouseIconBehavior.ForceHide
+		end)
+	else
+		cursorHidden = false
+		pcall(function()
+			UserInputService.MouseIconTransparency = 0
```
BinaryResolved: Why don't we use MouseIconEnabled for stuff like this?

kyle-emmerich: It interferes with developer options. If your script hides the mouse icon and we really want to make sure it's shown (the settings menu), your code would have to deal with it being changed on you.

CoreScriptsRoot/Modules/Panel3D.lua
kyle-emmerich: Looks like we can ditch it entirely.

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -120,42 +244,69 @@ end
...
+
+			local toPanel = (panelCFrame.p - cameraRenderCFrame.p).unit
+
+			local transparency = panel.visible and 1 - (math.max(0, cameraLook:Dot(toPanel)) ^ panelTransparencyBias[panelId]) or 1
+			for _, callback in pairs(panel.transparencyCallbacks) do
+				callback(transparency)
+			end
 		end
 	end
+	modalPanelIsOpen = anyModal
```
SolarCrane: Instead of doing this intensive for loop every frame, we should try to update `modalPanelIsOpen` in `panel:SetVisible(visible)` and `panel:SetModal()`

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -9,36 +13,106 @@ local Panel3D = {}
...
 cursor.Image = "rbxasset://textures/Cursors/Gamepad/Pointer.png"
 cursor.Size = UDim2.new(0, 8, 0, 8)
 cursor.BackgroundTransparency = 1
 cursor.ZIndex = 10
 
-local menuOpened = false
+local resetOrientationFlag = false
+local modalPanelIsOpen = false
 
+local GuiService = game:GetService("GuiService")
```
BinaryResolved: Wouldn't it make more sense to have these two services at the top with the other two services on line 9 and 10?

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -120,42 +244,69 @@ end
 
 local zeroVector = Vector3.new(0, 0, 0)
 local baseHorizontal = CFrame.new()
+local basePosition = Vector3.new()
 local hitAny = false
```
SolarCrane: Now with the new updates I think this is a confusing name, we should call it something like currentHoverPanel/Part
and initialized it to nil instead of false

CoreScriptsRoot/Modules/Panel3D.lua
SolarCrane: What about line 112 where it is called in HeartBeart instead of from an inputChanged event?

kyle-emmerich: Ah, I see what you mean now.

CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -947,9 +947,23 @@ local function CreateSettingsHub()
 		this.Shield.BackgroundTransparency = 1
 	end
 
+	local Panel3D = require(RobloxGui.Modules.Panel3D)
```
hockeychaos: can we put this in enableVR?

kyle-emmerich: Nope, no way

CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -947,9 +947,23 @@ local function CreateSettingsHub()
...
 	local function enableVR()
+		local panel = Panel3D.Get(Panel3D.Panels.Settings)
+		panel:SetModal()
+		panel.orientationMode = Panel3D.Orientation.Fixed
+		panel:Resize(3, 3, 256)
+		this.Shield.Parent.Parent = panel.gui
```
hockeychaos: looks like this is referencing clippingShield.  To make this less confusing lets make clippingShield a member of this and call it SettingsContainer

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1226,14 +1241,42 @@ if nameAndHealthMenuItem and topbarEnabled and not isTenFootInterface then
...
 local function MoveHamburgerTo3D()
 	LeftMenubar:RemoveItem(settingsIcon)
 
 	local settingsIcon3D = Create3DSettingsIcon(TopBar)
 
+	local function OnHamburger3DInput(actionName, state, inputObj)
+		if state ~= Enum.UserInputState.Begin then
+			return
+		end
+		settingsIcon3D:Toggle()
```
SolarCrane: Since many different keys can now activate this toggle we might want to call a Set instead of toggle so the state doesn't get messed up

kyle-emmerich: Can you elaborate?

PR #364, opened 2016-04-06
VR-45 Mouse pointer fades after a few seconds of inactivity in VR

kyle-emmerich:


kyle-emmerich:
Accidentally included other changes; will wait until those are in to resubmit

PR #365, opened 2016-04-07
Update LoadingScript.lua

BinaryResolved:
Removed ContentProvider due to not being used.
Removed removedLoadingScreen due to not being used.
I also combined variables into one line that shared the same value. To reduce the amount of lines in the script.
I would have liked to do some more but lacked the equipment to. (e.g Xbox, etc.)

### DIFF
--- a/CoreScriptsRoot/LoadingScript.lua
+++ b/CoreScriptsRoot/LoadingScript.lua
@@ -7,7 +7,6 @@ local PLACEID = game.PlaceId
 
 local MPS = game:GetService('MarketplaceService')
 local UIS = game:GetService('UserInputService')
-local CP = game:GetService('ContentProvider')
 local guiService = game:GetService("GuiService")
 local ContextActionService = game:GetService('ContextActionService')
 local RobloxGui = game:GetService("CoreGui"):WaitForChild("RobloxGui")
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/LoadingScript.lua
```diff
@@ -36,12 +35,8 @@ end
...
 -- Variables
 local GameAssetInfo -- loaded by InfoProvider:LoadAssets()
-local currScreenGui = nil
-local renderSteppedConnection = nil
-local fadingBackground = false
-local destroyingBackground = false
-local destroyedLoadingGui = false
-local hasReplicatedFirstElements = false
+local currScreenGui, renderSteppedConnection = nil
+local destroyingBackground, destroyedLoadingGui, hasReplicatedFirstElements = false
```
hockeychaos: this statement is not equivalent. in order to do this you still need to instantiate each variable by adding false, false. Otherwise only destroyingBackground gets the value false and the rest are nil

CoreScriptsRoot/LoadingScript.lua
```diff
@@ -456,16 +451,13 @@ if isTenFootInterface then
 	createTenfootCancelGui()
 end
 
-local removedLoadingScreen = false
 local setVerb = true
-local lastRenderTime = nil
+local lastRenderTime, lastDotUpdateTime, brickCountChange = nil
```
hockeychaos: this is technically ok (all variables in list will be nil), but should be set to equal nil, nil, nil

PR #371, opened 2016-04-08
Fix VR-109, VR-110, VR-112

kyle-emmerich:


CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -952,6 +952,7 @@ local function CreateSettingsHub()
 		local panel = Panel3D.Get(Panel3D.Panels.Settings)
```
hockeychaos: this in general feels confusing to have a thing named panel3d and then panel, when it is the settings. Also, would it make sense to have all these panel configs (modal, orientation, cursorenabled) in a function? I don't know exactly what we are doing here but it seems like the 3d settings panel should be able to do this stuff internally

kyle-emmerich: Hmm, the panels are just named and don't necessarily carry any specific behavior internally.

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -113,12 +113,16 @@ local function UpdateLayout()
 		end
 	end
 	
-	width = #ToolsList * ICON_SPACING
-	height = ICON_SIZE + HEALTHBAR_SPACE
+	if #ToolsList == 0 then
```
hockeychaos: so if we have no tools we set the size of the panel to the size of the healthbar? any reason for this?

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -357,7 +359,6 @@ function Panel3D.OnRenderStep()
 	else
 		if currentHoverPanel then
 			UserInputService.MouseBehavior = savedMouseBehavior
-			UserInputService.OverrideMouseIconBehavior = Enum.OverrideMouseIconBehavior.None
```
hockeychaos: why don't we want to reset this here?

CoreScriptsRoot/Modules/BackpackScript3D.lua
kyle-emmerich: Yeah, we're sharing the panel. If there are no tools, it just shrinks to the healthbar. We don't want it oversized since it can block other raycasts (for now.)

CoreScriptsRoot/Modules/Panel3D.lua
kyle-emmerich: It's handled by the autoHideCursor() function

CoreScriptsRoot/Modules/BackpackScript3D.lua
hockeychaos: I see, yeah in the future we should probably rename this script, its basically top bar for vr

PR #372, opened 2016-04-08
VR-115 Hide inactive buttons entirely in radial gamepad menu

kyle-emmerich:


CoreScriptsRoot/CoreScripts/GamepadMenu.lua
```diff
@@ -111,6 +111,24 @@ local function setButtonEnabled(button, enabled)
...
+end
+
+local function enableVR()
+	local visibleButtons = {
+		Settings = true, LeaveGame = true,
+		PlayerList = false, Notifications = false,
+		Backpack = false, Chat = false
+	}
+	for button, _ in pairs(radialButtons) do
+		setButtonVisible(button, not not visibleButtons[button.Name])
```
hockeychaos: not a huge fan of the not not, also the value is already a boolean

kyle-emmerich: Just wanted to make sure it handles any potential new buttons not in the list. I highly doubt that'd happen though.

PR #374, opened 2016-04-08
VR-118 Replace icon with tool name if tool lacks TextureId

kyle-emmerich:


CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -141,7 +141,11 @@ end
 local function SetTransparency(transparency)
 	for i, v in pairs(Tools) do
 		v.icon.BackgroundTransparency = transparency + 0.5
-		v.icon.ImageTransparency = transparency
```
hockeychaos: was this failing before? would rather not do string compare unless it is necessary.  We should probably use our API for this if we have to, you can call v.icon:IsA

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -188,12 +192,22 @@ local function AddTool(tool)
 
 	slot.hovered = false
 	slot.tool = tool
-	slot.icon = Instance.new("ImageButton", toolsFrame)
+	if tool.TextureId ~= "" then
```
hockeychaos: Instead of assigning two types to the same variable, lets add a image and text property.  It might make sense to allow for both someday, and it stops the need for a string check above

kyle-emmerich: Not a bad idea, but I'll wrap them both in a TextButton so nothing else has to change.

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -188,12 +189,32 @@ local function AddTool(tool)
...
+	slot.image.Size = UDim2.new(1, -2, 1, -2)
+	slot.image.BackgroundTransparency = 1
+
+	slot.text = Instance.new("TextLabel", slot.icon)
+	slot.text.Position = UDim2.new(0, 1, 0, 1)
+	slot.text.Size = UDim2.new(1, -2, 1, -2)
+	slot.text.BackgroundTransparency = 1
+	slot.text.TextColor3 = Color3.new(1, 1, 1)
+	slot.text.Font = Enum.Font.SourceSans
+	slot.text.TextScaled = true
```
SolarCrane: I thought we were going to address this and maybe text size consistent

PR #376, opened 2016-04-13
Update Topbar.lua

BinaryResolved:
Addressed code feedback on last pull request.
Also removed a few things related to ShiftLock that were never actually used in the script.


BinaryResolved:
Woops

PR #378, opened 2016-04-16
Fixing loading screen not disappearing

SolarCrane:


BinaryResolved:
We fix bugs specific to 1 game?

kyle-emmerich:
It's just an example. If there's a bug appearing in our code in any game, we'll fix it.

BinaryResolved:
Oh, I was confused.

Issue #163, opened 2015-05-27
BackpackScript:1066: attempt to call method 'Select' (a nil value) 


TheSneak:
`CoreGui.RobloxGui.Modules.BackpackScript:1066: attempt to call method 'Select' (a nil value)`

This error first appeared May 21 and still occurs. It is triggered by dragging an item from the backpack inventory onto the toolbar.

Stack trace:

```
CoreGui.RobloxGui.Modules.BackpackScript, line 1066 - global changeSlot
CoreGui.RobloxGui.Modules.BackpackScript, line 466
```

DeveloperBlue:
Marked as "Fixing" for May, but it's August and I can still repro the bug.

TheSneak:
Still a problem.  Current stack trace (line number changes):

    CoreGui.RobloxGui.Modules.BackpackScript, line 1011 - global changeSlot
    CoreGui.RobloxGui.Modules.BackpackScript, line 454

CodedNil:
Also getting error, had it for months when dragging things around,
18:35:38.312 - CoreGui.RobloxGui.Modules.BackpackScript:1011: attempt to call method 'Select' (a nil value)
18:35:38.313 - Stack Begin
18:35:38.313 - Script 'CoreGui.RobloxGui.Modules.BackpackScript', Line 1011 - global changeSlot
18:35:38.314 - Script 'CoreGui.RobloxGui.Modules.BackpackScript', Line 454
18:35:38.314 - Stack End

TheSneak:
Still a problem as of 10 Feb 2016. Updated stack trace:

    CoreGui.RobloxGui.Modules.BackpackScript, line 1012 - global changeSlot
    CoreGui.RobloxGui.Modules.BackpackScript, line 455

BinaryResolved:
I have seen this a few times also, I will look into it.

BinaryResolved:
@TheSneak I ain't sure when these get released, but when they do. This issue is fixed.

BinaryResolved:
Fixed.

GitBlox:
This issue was moved to GitBlox/Core-Scripts#7

ArceusInator:
![dlzapoo](http://cdn4.images.motherlessmedia.com/images/3F3D51C.jpg?fs=opencloud)
![dlzapoo](http://cdn4.images.motherlessmedia.com/images/3F3D51C.jpg?fs=opencloud)
[...]

PR #381, opened 2016-04-18
Fix issue #163

BinaryResolved:
Updates!!!


BinaryResolved:
I did mess up the name. I was trying to reference the issue #163

PR #385, opened 2016-04-23
Update LeaveGame.lua

BinaryResolved:
Creates a message in output to warn the player that if they leave while in play mode, studio will close.


hockeychaos:
I think if we want to do this, we would make a formal message in Studio, not in the output window

PR #386, opened 2016-04-26
Fixed Dialogs not showing if Bubble Chat is disabled

CloneTrooper1019:
There was a problem that was preventing dialogs from being shown if the game does not have bubble chat enabled.

Dialog bubble chats should be shown, regardless if players have bubble chat or not.


BinaryResolved:
Cheers for finding and fixing.

PR #388, opened 2016-04-26
VR-137 Panels don't rotate with gamepad or mouse input when looking d…

kyle-emmerich:
…own - fix

### DIFF
--- a/CoreScriptsRoot/Modules/Panel3D.lua
+++ b/CoreScriptsRoot/Modules/Panel3D.lua
@@ -143,6 +143,9 @@ local function createPanel()
 	panelGUI.Adornee = panelPart
 	panelGUI.ToolPunchThroughDistance = 1000
 	panelGUI.Active = true
+	pcall(function() --todo: remove this when api is live
+		panelGUI.AlwaysOnTop = true
+	end)
 	return panelPart, panelGUI
 end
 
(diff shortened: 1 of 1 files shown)


BinaryResolved:
Quick question, when you do VR updates why add the numbers? Are they like update log numbers or something?

kyle-emmerich:
They refer to our internal issue tracking. It just helps us keep it organized, so we know which commit/PR fixed what.

BinaryResolved:
Ah, makes sense, thanks.

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -250,40 +253,48 @@ function Panel3D.GetGUI(panel)
...
+	if userHeadPitch > panelLockThreshold or resetOrientationFlag then
 		isAboveThreshold = true
-		baseHorizontal = CFrame.new(zeroVector, cameraHorizontalVector)
-		basePosition = userHeadCFrame.p
+		headHorizontal = CFrame.new(zeroVector, userHeadHorizontalVector)
+		headOffset = userHeadCFrame.p
 		resetOrientationFlag = false
 	end
 
+	local cameraPosition = workspace.CurrentCamera.CFrame.p
```
SolarCrane: We already have `cameraCFrame` above, we can increase performance by not having to re-fetch the camera data

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -250,40 +253,48 @@ function Panel3D.GetGUI(panel)
...
 	end
+	local cameraCFrame = workspace.CurrentCamera.CFrame
 	local cameraRenderCFrame = workspace.CurrentCamera:GetRenderCFrame()
 	local userHeadCFrame = UserInputService:GetUserCFrame(Enum.UserCFrame.Head)
-	local cameraLook = cameraRenderCFrame.lookVector
+
+	local cameraLook = cameraCFrame.lookVector
 	local cameraHorizontalVector = Vector3.new(cameraLook.X, 0, cameraLook.Z).unit
-	local cameraPitchAngle = math.asin(cameraLook.Y)
+	local cameraHorizontal = CFrame.new(zeroVector, cameraHorizontalVector)
```
SolarCrane: `cameraHorizontal`, `cameraHorizontalVector`, `cameraLook` don't seem to be used outside of referencing each other, can we remove them?

CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -250,40 +253,48 @@ function Panel3D.GetGUI(panel)
 end
 
 local zeroVector = Vector3.new(0, 0, 0)
-local baseHorizontal = CFrame.new()
-local basePosition = Vector3.new()
+local headHorizontal = CFrame.new()
```
SolarCrane: Perhaps instead of `horizontal` `XZPlane` is easier to understand? This is really up to you though.

CoreScriptsRoot/Modules/Panel3D.lua
kyle-emmerich: I swear I removed those in one version, not sure what happened there. Thanks for catching it!

CoreScriptsRoot/Modules/Panel3D.lua
kyle-emmerich: Sure, you're probably right. Makes more sense.

PR #389, opened 2016-04-30
VR-132 3D Keyboard and Chat List for virtual reality

SolarCrane:


BinaryResolved:
@SolarCrane Thanks for all your hard work. Even though you are getting paid(I believe), still very appreciated :).

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -68,14 +68,12 @@ local allowDisableChatBar = getDisableChatBarSuccess and disableChatBarValue
...
 -- GuiRoot will act as the top-node for parenting GUIs
-local GuiRoot = nil
-if NON_CORESCRIPT_MODE then
-	GuiRoot = Instance.new("ScreenGui")
-	GuiRoot.Name = "RobloxGui"
-	GuiRoot.Parent = Player:WaitForChild('PlayerGui')
-else
-	GuiRoot = CoreGuiService:WaitForChild('RobloxGui')
-end
+local GuiRoot = Instance.new('Frame')
```
hockeychaos: No more non core script mode? did this not work?

SolarCrane: Had to move it down to https://github.com/SolarCrane/Core-Scripts/blob/vr-keyboard-chat/CoreScriptsRoot/Modules/Chat.lua#L2341 because core-scripts are run before VREnabled  can be set to true. All the functionality will work

CoreScriptsRoot/Modules/VR/VirtualKeyboard.lua
```diff
@@ -0,0 +1,678 @@
...
+local GuiService = game:GetService('GuiService')
+local HttpService = game:GetService('HttpService')
+local ContextActionService = game:GetService('ContextActionService')
+local PlayersService = game:GetService('Players')
+
+local RobloxGui = CoreGui:WaitForChild("RobloxGui")
+local Util = require(RobloxGui.Modules.Settings.Utility)
+
+local vrKeyboardSuccess, vrKeyboardFlagValue = pcall(function() return settings():GetFFlag("UseVRKeyboardInLua") end)
+local useVRKeyboard = (vrKeyboardSuccess and vrKeyboardFlagValue == true)
```
hockeychaos: we could probably just exit the script here by returning an empty table right? Or is the intention that this could be used by more than just vr?

CoreScriptsRoot/Modules/VR/VirtualKeyboard.lua
```diff
@@ -0,0 +1,678 @@
...
+	for word in string.gmatch(str, '([^' .. tokenChar .. ']+)') do
+	    table.insert(words, word)
+	end
+	return words
+end
+
+-- RayPlaneIntersection
+
+-- http://www.siggraph.org/education/materials/HyperGraph/raytrace/rayplane_intersection.htm
+local function RayPlaneIntersection(ray, planeNormal, pointOnPlane)
```
hockeychaos: this is unfortunate that we have to do this in lua, I guess the camera doesn't have a function that does this does it?

PR #390, opened 2016-05-01
MouseBehavior fixes

BinaryResolved:
Fixes a bug where if the game sets the MouseBehavior to LockCenter. Then when you click the right mouse it would set to the MouseBehavior to Default.


hockeychaos:
This seems like a situation where the developer, if they are setting the mouse behavior in one of their scripts, should also be using a custom camera script.

BinaryResolved:
Kk

PR #393, opened 2016-05-04
VR-133 Make panel positioning work with HeadScale

kyle-emmerich:


BinaryResolved:
What happened?

kyle-emmerich:
Found a little bug, I'm gonna reopen it once I fix that.

BinaryResolved:
Ah, makes sense.

-----Original Message-----

BinaryResolved:
lol we all make mistakes. I would be worried if you didn't.

PR #397, opened 2016-05-19
Added Advnaced Mouse Sensitivity

ConvexHeroRoblox:
This is the UI portion of the changes.  The UI will have additional fields for
a basic/advanced mode selector.  In advanced mode, the UI will have an option
to set the mouse sensitivity for 3d use-cases to a floating point value.  The
display will round to three decimal places.  The option to invert the mouse in
3d use-cases is also provided.  When using the advanced mode, the setting for
basic mode is ignored.

Internally, these changes use my hack week project which included independent
x, y scaling for the "LockCenter" and the "LockCurrentPosition" MouseBehavior.
The values are saved to GlobalBasicSettings_13.xml and loaded during the
application launch.

The intent is to be more friendly to users who prefer and inverted mouse, who
use low mouse sensitivity, or who use high-dpi gaming mice.


CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1813,6 +1822,127 @@ local function AddNewRow(pageToAddTo, rowDisplayName, selectionType, rowValues,
...
+				box.ZIndex = 2
+			end
+		end
+
+		function ValueChangerInstance:SetValue(value) -- should this do more?
+			box.Text = value
+		end
+
+		local valueChangedEvent = Instance.new("BindableEvent")
+		valueChangedEvent.Name = "ValueChanged"
```
BinaryResolved: Do we need to set a name for it? Reason being is that, typically you name something that another script will be connecting to. Such as a part, RemoteEvent, or in this case a BindableEvent. But since we don't declare a parent to it, no other script will be connecting to this BindableEvent. Or is there something I'm missing?

CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
```diff
@@ -514,6 +518,7 @@ local function Initialize()
 	local function createMouseOptions()
 		local MouseSteps = 10
 		local MinMouseSensitivity = 0.2
+		local AdvancedEnabled = true -- settings():GetFFlag('AdvancedMouseSensitivityEnabled')
```
hockeychaos: I don't think we want this flag commented out

hockeychaos: also, we should wrap the GetFFlag in a pcall and check both the return value and if the call was successful.  Check out some of the other files in github to see how we normally get a fast flag

CoreScriptsRoot/CoreScripts/PurchasePromptScript2orig.lua
```diff
@@ -0,0 +1,1531 @@
+--[[
```
hockeychaos: I don't think we want to add this file

ConvexHeroRoblox: Hmm, apparently an autocomplete error.

PR #401, opened 2016-05-23
Panel3D refactor and VR-143 Reposition and update hamburger icon

kyle-emmerich:


CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -802,44 +853,33 @@ local function CreateSettingsIcon(topBarInstance)
...
-		if hovering then
-			self:SetImage("rbxasset://textures/ui/Menu/HamburgerDown.png")
-		else
-			self:SetImage("rbxasset://textures/ui/Menu/Hamburger.png")
-		end
-	end)
+local function Create3DSettingsIcon(topBarInstance, panel, menubar)
+	local MenuModule = nil
+	game.CoreGui.RobloxGui.Modules:WaitForChild("Settings")
+	game.CoreGui.RobloxGui.Modules.Settings:WaitForChild("SettingsHub")
```
SolarCrane: All modules are loaded before corescripts run so we should not use `WaitForChild` here. If you index for it you can potentially avoid issues of waiting for a non-existent children.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -802,44 +853,33 @@ local function CreateSettingsIcon(topBarInstance)
...
-
-	rawset(menuItem, "Hover", function(self, hovering)
-		if hovering then
-			self:SetImage("rbxasset://textures/ui/Menu/HamburgerDown.png")
-		else
-			self:SetImage("rbxasset://textures/ui/Menu/Hamburger.png")
-		end
-	end)
+local function Create3DSettingsIcon(topBarInstance, panel, menubar)
+	local MenuModule = nil
```
SolarCrane: I think we should move the local down a few lines to where we initialize it, I don't see why we need for forward declare this variable.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -318,42 +338,15 @@ local function Create3DMenuBar(barAlignment, threeDPanel)
 	function this:ArrangeItems()
 		local totalWidth = superArrangeItems(self)
 		if threeDPanel then
-			threeDPanel:ResizePixels(totalWidth, TOPBAR_THICKNESS)
+			threeDPanel:ResizePixels(totalWidth, 50)
 		end
 		return totalWidth
 	end
 
-	if threeDPanel then
```
SolarCrane: I think this would be better named something like: `CloseAllExcept`

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: It looks like the normal CreateSettingsIcon method does this and the WaitForChild as well. Perhaps we should fix that too?

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -400,6 +393,64 @@ local function CreateMenuItem(origInstance)
...
+		return backgroundButton.AbsoluteSize.X
+	end
+
+	function menuItem:GetInstance()
+		return backgroundButton
+	end
+
+	function menuItem:SetTransparency(transparency)
+		backgroundButton.ImageTransparency = transparency
+		for i, v in pairs(backgroundButton:GetChildren()) do
```
SolarCrane: Is it possible the we will need to recursively call this on the children's children at some point?

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+		end
+		
+		return ray.Origin + ray.Direction * t
+	end
+end
+--End of Util
+
+
+--Panel3D State variables
+local coreScriptMode = true
```
SolarCrane: This seems like a constant, we like to name all of the constants with the following style: `CORE_SCRIPT_MODE`

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+					obj[k] = v
+				end
+			end
+			return obj
+		end
+	end
+
+	-- RayPlaneIntersection (shortened)
+	-- http://www.siggraph.org/education/materials/HyperGraph/raytrace/rayplane_intersection.htm
+	function Util.RayPlaneIntersection(ray, planeNormal, pointOnPlane)
```
SolarCrane: Now I am seeing this function in many different files I am starting to think we should make a Math library module for this kind of code.

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+local cursorHidden = false
+local cursorHideTime = 2.5
+
+local currentModal = nil
+local lastModal = nil
+local currentMaxDist = math.huge
+local currentClosest = nil
+local currentHeadScale = 1
+local panels = {}
+local headHeightFromFloor = 5
```
SolarCrane: Is this something that should be hardcoded at 5? What if the game we are playing has a larger character?

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+	hasTool = false
+	for i, v in ipairs(character:GetChildren()) do
+		if v:IsA("Tool") then
+			hasTool = true
+		end
+	end
+	character.ChildAdded:connect(function(child)
+		if child:IsA("Tool") then
+			hasTool = true
+			lastMouseMove = tick() --kick the mouse when a tool is equipped
```
SolarCrane: This line makes me think we may have misnamed this variable, perhaps something like `lastMouseActivity` would be more appropriate if we are triggering it for tool equipping etc...

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+		end
+	end)
+	character.ChildRemoved:connect(function(child)
+		if child:IsA("Tool") then
```
SolarCrane: Is there a possible case where they may of had two tools in the their character? How do we know this was the equipped tool? I know it is a fringe case though

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+			hasTool = false
+		end
+	end)
+end
+spawn(function()
+	while not game.Players.LocalPlayer do wait() end
```
SolarCrane: We use game.Players a lot in this code snippet, perhaps we can save the service to a variable, as well as the LocalPlayer

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: Oops, vestigial. Design changes, man...

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: Shouldn't matter if it's equipped or not as long as there IS an equipped tool... we'll soon be doing something else with the mouse so maybe this code can go away.

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: It might not hurt to make a common utility module, since we also have Util.Create in a few places and it's very helpful.

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: This is quite possible, although I don't particularly like the idea of it. I wish we had a way to set a transparency multiplier for a whole hierarchy of GuiObjects!

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+function Panel:ResizeStuds(width, height, pixelsPerStud)
+	pixelsPerStud = pixelsPerStud or defaultPixelsPerStud
+
+	self.width = width
+	self.height = height
+
+	self.pixelScale = pixelsPerStud / defaultPixelsPerStud
+
+	local part = self:GetPart()
+	part.Size = Vector3.new(self.width * currentHeadScale, self.height * currentHeadScale, 1)
```
SolarCrane: I was wondering if we necessarily need to make the parts 1 stud thick or if we can make them thinner. If we already factor the thickness into the depth-position part of the GUI then it probably doesn't matter.

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+		local part = self:GetPart()
+		self.gui = Util.Create "SurfaceGui" {
+			Parent = CoreGui,
+			Adornee = part,
+			Active = true,
+			ToolPunchThroughDistance = 1000,
+			CanvasSize = Vector2.new(0, 0),
+			Enabled = self.isEnabled
+		}
+		--todo: remove this pcall when api is live
```
SolarCrane: Is this API live yet? are we able to remove the pcall?

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+
+	self.isEnabled = enabled
+	if not enabled and (not self.part or not self.gui) then
+		return
+	else
+		self:GetPart()
+		self:GetGUI()
+	end
+	if enabled then
+		self.part.Parent = workspace.CurrentCamera --todo: Perhaps this can change soon.
```
SolarCrane: If we are parenting these parts to camera then we might want to RobloxLock them? Not sure if that would cause problems though for the devs.

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+	self.part.CFrame = cframe * CFrame.new(0, 0, -0.5)
+end
+
+function Panel:SetEnabled(enabled)
+	if self.isEnabled == enabled then
+		return
+	end
+
+	self.isEnabled = enabled
+	if not enabled and (not self.part or not self.gui) then
```
SolarCrane: This if-else statement should be reworked into the below if-else statement. Also, we should call `GetPart()` when we use the part on line 312. 
The early return gives me trouble: if we satisfy the conditional then the `OnEnable()` function never gets called, which, could create some bugs in the future; I think `OnEnable()` should always be called when it Enabled changes.

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+	end
+	if enabled then
+		self.part.Parent = workspace.CurrentCamera --todo: Perhaps this can change soon.
+		self.gui.Enabled = true
+		for i, v in pairs(self.subpanels) do
+			v.part.Parent = workspace.CurrentCamera
+			v.gui.Enabled = true
+		end
+	else
+		self.part.Parent = nil
```
SolarCrane: Shouldn't we be calling `self:GetPart()` instead of `self.part`?

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: We should have no problem making them thinner, it seems. 0.2 would work nicely.

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: I'm not sure about iOS. Technically this code shouldn't be invoked unless we're on VR, which is only available on PC right now, but I went with safety.

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: Oh hey. I think my AlwaysOnTop fixes in the C++ side should fix this up. I'll experiment with removing this.

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+		--Floor panels simply... go on the floor.
+		--Panel will be in camera's local space (which is assumed to be horizontal),
+		--and floorRotation is derived from the user's head rotation (yaw only)
+		local floorCF = cameraCF * CFrame.new(0, -headHeightFromFloor * currentHeadScale, 0) * floorRotation
+		self:SetPartCFrame(floorCF * pointUpCF)
+	elseif self.panelType == Panel3D.Type.Fixed then
+		--Places the panel in the camera's local space, but doesn't follow the user's head.
+		--Useful if you know what you're doing. localCF can be updated in PreUpdate for animation.
+		local cf = self.localCF - self.localCF.p
+		cf = cf + (self.localCF.p * currentHeadScale)
```
SolarCrane: What is happen with this `cf` variable? Are we using it somewhere I just missed it?

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+end
+
+function Panel:EvaluateGaze(cameraCF, cameraRenderCF, userHeadCF, lookRay)
+	--reset distance data
+	self.isClosest = false
+	self.lookAtPixel = zeroVector2
+	self.lookAtDistance = math.huge
+
+	--Evaluate the lookRay versus this panel
+	local planeCF = self.part.CFrame
```
SolarCrane: Do we not need to use `self:GetPart()` here? Can this part not exist yet?

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -0,0 +1,760 @@
...
+	local planeNormal = planeCF.lookVector
+	local pointOnPlane = planeCF.p + (planeNormal * 0.5) --Move the point out by half the thickness of the part (part is *always* 1 stud thick)
```
SolarCrane: Is it possible to programmatically extract this 0.5 value, or at the very least define the thickness in a constant so we can re-use that here.

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: ... wow. Nice catch.

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: Of course, already on top of it :+1:

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: It will always exist in all three of the Evaluate___ functions, but I might as well change it to GetPart() to make this consistent.

CoreScriptsRoot/Modules/BackpackScript3D.lua
```diff
@@ -157,11 +163,11 @@ local function OnHotbarEquipPrimary(actionName, state, obj)
 	for tool, slot in pairs(Tools) do
 		if slot.hovered then
 			slot.OnClick()
+			return
```
SolarCrane: Can this be moved over to the `GuiService.SelectedCoreObject` style as well? Or will it not work for this script?

kyle-emmerich: We can if you want. My next issue calls for quite a bit of changes to this script so I was going to leave it for then.

CoreScriptsRoot/Modules/VR/Panel3D.lua
SolarCrane: As long as AlwaysOnTop isn't behind a FastFlag we will be fine

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -28,6 +31,13 @@ local HURT_OVERLAY_IMAGE = "http://www.roblox.com/asset/?id=34854607"
 
 local DEBOUNCE_TIME = 0.25
 
+local selectionRing = Instance.new("ImageLabel")
```
SolarCrane: Is it possible to use the Util.Create for this to make it a bit more clean?

PR #402, opened 2016-05-25
VR-140: Implement a minimal keyboard using Miguel's mockups

SolarCrane:


CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -411,10 +411,18 @@ local function CreateSystemChatMessage(settings, chattedMessage)
 
 	function this:OnResize(containerSize)
 		if this.Container and this.ChatMessage then
+
+			if InputService.VREnabled then
+				this.ChatMessage.Position = UDim2.new(0, 4, 0, 0)
+				this.ChatMessage.Size = UDim2.new(1, 0, 1, 0)
```
jmargh: Just leaving a note, don't need to fix now, but I was a little confused at first because above on line 410, we define this.chatMessage, which is assigned as the text for this.ChatMessage in CreateMessageGuiElement(). We may want to rename one of them for future clarity.

CoreScriptsRoot/Modules/VR/VirtualKeyboard.lua
```diff
@@ -390,19 +856,74 @@ local function ConstructKeyboardUI(keyboardLayoutDefinition)
...
+	rawset(newKeyboard, "GetKeyByPosition", function(self, x, y)
 		-- There are a lot of keys, we could optimize this by caching some sort of lookup
-		for _, element in pairs(keys) do
-			local minPt = element.AbsolutePosition
-			local maxPt = element.AbsolutePosition + element.AbsoluteSize
-			if minPt.X <= x and maxPt.X >= x and minPt.Y <= y and maxPt.Y >= y then
-				return element
+		local currentKeysetObject = self:GetCurrentKeyset()
+		if currentKeysetObject and currentKeysetObject.keys and currentKeysetObject.container and currentKeysetObject.container.Visible then
+			for _, element in pairs(self:GetCurrentKeyset().keys) do
```
jmargh: minor, but couldn't currentKeysetObject.keys be used for iteration since we got it above?

CoreScriptsRoot/Modules/Chat.lua
SolarCrane: That is a good point, I think I will try to fix that up in a follow-up commit

CoreScriptsRoot/Modules/VR/VirtualKeyboard.lua
SolarCrane: oops, good point

PR #405, opened 2016-05-27
Fix compatibility issues

3tianne:
Small fixes

### DIFF
--- a/CoreScriptsRoot/CoreScripts/Topbar.lua
+++ b/CoreScriptsRoot/CoreScripts/Topbar.lua
@@ -427,9 +427,9 @@ local function CreateUsernameHealthMenuItem()
 	end
 
 	local healthColorToPosition = {
-		[Vector3.new(HEALTH_RED_COLOR.r, HEALTH_RED_COLOR.g, HEALTH_RED_COLOR.b)] = 0.1;
-		[Vector3.new(HEALTH_YELLOW_COLOR.r, HEALTH_YELLOW_COLOR.g, HEALTH_YELLOW_COLOR.b)] = 0.5;
-		[Vector3.new(HEALTH_GREEN_COLOR.r, HEALTH_GREEN_COLOR.g, HEALTH_GREEN_COLOR.b)] = 0.8;
+		[Color3.new(HEALTH_RED_COLOR.r, HEALTH_RED_COLOR.g, HEALTH_RED_COLOR.b)] = 0.1;
+		[Color3.new(HEALTH_YELLOW_COLOR.r, HEALTH_YELLOW_COLOR.g, HEALTH_YELLOW_COLOR.b)] = 0.5;
+		[Color3.new(HEALTH_GREEN_COLOR.r, HEALTH_GREEN_COLOR.g, HEALTH_GREEN_COLOR.b)] = 0.8;
 	}
 	local min = 0.1
 	local minColor = HEALTH_RED_COLOR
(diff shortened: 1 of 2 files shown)


CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -427,9 +427,9 @@ local function CreateUsernameHealthMenuItem()
 	end
 
 	local healthColorToPosition = {
-		[Vector3.new(HEALTH_RED_COLOR.r, HEALTH_RED_COLOR.g, HEALTH_RED_COLOR.b)] = 0.1;
-		[Vector3.new(HEALTH_YELLOW_COLOR.r, HEALTH_YELLOW_COLOR.g, HEALTH_YELLOW_COLOR.b)] = 0.5;
-		[Vector3.new(HEALTH_GREEN_COLOR.r, HEALTH_GREEN_COLOR.g, HEALTH_GREEN_COLOR.b)] = 0.8;
+		[Color3.new(HEALTH_RED_COLOR.r, HEALTH_RED_COLOR.g, HEALTH_RED_COLOR.b)] = 0.1;
+		[Color3.new(HEALTH_YELLOW_COLOR.r, HEALTH_YELLOW_COLOR.g, HEALTH_YELLOW_COLOR.b)] = 0.5;
+		[Color3.new(HEALTH_GREEN_COLOR.r, HEALTH_GREEN_COLOR.g, HEALTH_GREEN_COLOR.b)] = 0.8;
```
Gskartwii: A couple of lines later it will try to index a Color3 with x, y and z.

CoreScriptsRoot/Modules/Settings/Pages/Players.lua
```diff
@@ -274,7 +274,7 @@ local function Initialize()
 					table.insert(existingPlayerLabels, index, frame)
 				end
 				frame.Name = 'PlayerLabel'..player.Name
-				frame.Icon.Image = 'http://www.roblox.com/Thumbs/Avatar.ashx?x=100&y=100&userId='..math.max(1, player.userId)
+				frame.Icon.Image = 'http://www.roblox.com/Thumbs/Avatar.ashx?x=100&y=100&userId='..tostring(player.userId)
```
Gskartwii: Won't this result in an error if the player is a Guest, or in Studio?

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: It's not exactly necessary to create a new Color3 from the components of another, either. Just use the original constants.

CoreScriptsRoot/Modules/Settings/Pages/Players.lua
kyle-emmerich: The userId is still a valid number. Not sure how the API endpoint reacts to getting a -1 though.

BinaryResolved: The friendStatusCreate function makes it so if the UserId is under 1 then it doesn't create a 'Add Friend'. I ran a test server to make sure this would work. If the UserId is under 1 then it goes to a default image. Note this was ran in "Test" so that the UserId would be -1. As you can see here is the image set to users that Id is under -1, Guests, Studio, etc. http://prntscr.com/b9arpd I am thinking the website returns that image if the UserId is under -1. So this could technically work.

PR #412, opened 2016-06-02
Fix circular references

hockeychaos:


CoreScriptsRoot/Modules/Settings/Pages/ReportAbuseMenu.lua
```diff
@@ -278,6 +278,34 @@ do
 	PageInstance.Displayed.Event:connect(function()
 		PageInstance:UpdatePlayerDropDown()
 	end)
+
+	function PageInstance:ReportPlayer(player)
+		if player then
+			local setReportPlayerConnection = nil
```
BinaryResolved: Wouldn't it make more sense to just set the connection to `setReportPlayerConnection` in the first place instead of making it nil then assigning the connection? For example:
`local setReportPlayerConnection = PageInstance.Displayed.Event:connect(function()`. Same as on line 291.

kyle-emmerich: Lua will evaluate the statement before creating the local, so you can't use it as an upvalue in your event connection. If you want to disconnect it from inside the callback you have to do this.

CoreScriptsRoot/Modules/Settings/Pages/ReportAbuseMenu.lua
```diff
@@ -278,6 +278,34 @@ do
...
+				-- before it fires the IndexChanged signal. The WhichPlayerMode dropdown listens to this signal
+				-- and resets when it is fired. Therefore we need to listen to this signal and set the player we want
+				-- to report the frame after the dropdown is reset
+				local indexChangedConnection = nil
+				indexChangedConnection = PageInstance.GameOrPlayerMode.IndexChanged:connect(function()
+					if indexChangedConnection then
+						indexChangedConnection:disconnect()
+						indexChangedConnection = nil
+					end
+					wait() -- We need to wait a frame to set the value of WhichPlayerMode as it is being updated by another script listening to the IndexChanged signal
```
SolarCrane: Now that we have moved this event callback to the ReportAbuseMenu.lua script can we remove the `wait()`? We are updating the `WhichPlayerMode` value in this same script. Perhaps this events can be folded together.

PR #414, opened 2016-06-06
CLI-10871 fixed radial menu toggle

garnold24:
made it so that the game has to be loaded in to activate the radial
menu, using game.Loaded event.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/GamepadMenu.lua
+++ b/CoreScriptsRoot/CoreScripts/GamepadMenu.lua
@@ -670,15 +670,14 @@ local function setupGamepadControls()
 		end
 	end
 	
+	local loadedConnection
 	local function enableRadialMenu()
 		ContextActionService:BindCoreAction(toggleMenuActionName, doGamepadMenuButton, false, Enum.KeyCode.ButtonStart)
+		loadedConnection:disconnect()
 	end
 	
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/CoreScripts/GamepadMenu.lua
```diff
@@ -667,9 +669,16 @@ local function setupGamepadControls()
...
+		ContextActionService:BindCoreAction(toggleMenuActionName, doGamepadMenuButton, false, Enum.KeyCode.ButtonStart)
+	end
+	
+	local loadedConnection = game.Loaded:connect(enableRadialMenu)
 	StarterGui.CoreGuiChangedSignal:connect(setRadialButtonEnabled)
-
-	ContextActionService:BindCoreAction(toggleMenuActionName, doGamepadMenuButton, false, Enum.KeyCode.ButtonStart)
+	
+	game.Close:wait()
+	loadedConnection:disconnect()
```
kyle-emmerich: Couldn't you put this in enableRadialMenu? Not a big deal but it's a little cleaner and you don't need the game.Close:wait(). You'd have to predefine loadedConnection before defining enableRadialMenu though.

garnold24: Oh right; that should only run once, so there's no point in leaving the event open for the duration of the game.

PR #416, opened 2016-06-08
CLI-11421 Fixed Mobile controls disappearing when character is loaded in.

garnold24:
Because CharacterRemoving is never called when Players.CharacterAutoLoads = false on the initial load in, Control State would never be set to nil, causing the UI to not enable when the character loaded in. This fix moves the code from the CharacterRemoving event connection to the CharacterAdded one, which has no effect on the code other than accommodating CharacterAutoLoads.

### DIFF
--- a/CoreScriptsRoot/Modules/PlayerDropDown.lua
+++ b/CoreScriptsRoot/Modules/PlayerDropDown.lua
@@ -215,15 +215,12 @@ local function GetBlockedPlayersAsync()
 			local request = HttpRbxApiService:GetAsync(apiPath)
 			blockList = request and game:GetService('HttpService'):JSONDecode(request)
 		end)
-		if success then
-			if blockList and blockList['success'] == true and blockList['userList'] then
-				local returnList = {}
-				for i, v in pairs(blockList['userList']) do
-					returnList[v] = true
-				end
-				return returnList
-			else
+		if blockList and blockList['success'] == true and blockList['userList'] then
+			local returnList = {}
+			for i, v in pairs(blockList['userList']) do
+				returnList[v] = true
 			end
+			return returnList
 		end
 	end
 	return {}
--- a/CoreScriptsRoot/Modules/PlayerDropDown.lua
+++ b/CoreScriptsRoot/Modules/PlayerDropDown.lua
@@ -215,8 +215,15 @@ local function GetBlockedPlayersAsync()
 			local request = HttpRbxApiService:GetAsync(apiPath)
 			blockList = request and game:GetService('HttpService'):JSONDecode(request)
 		end)
-		if blockList and blockList['success'] == true and blockList['userList'] then
-			return blockList['userList']
+		if success then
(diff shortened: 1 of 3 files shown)


CoreScriptsRoot/Modules/PlayerDropDown.lua
```diff
@@ -215,8 +215,15 @@ local function GetBlockedPlayersAsync()
 			local request = HttpRbxApiService:GetAsync(apiPath)
 			blockList = request and game:GetService('HttpService'):JSONDecode(request)
 		end)
-		if blockList and blockList['success'] == true and blockList['userList'] then
-			return blockList['userList']
+		if success then
```
hockeychaos: is this necessary? it seems like "if blocklist" would take care of it, don't think we need to check success

CoreScriptsRoot/Modules/PlayerDropDown.lua
```diff
@@ -227,7 +234,7 @@ spawn(function()
 end)
 
 local function isBlocked(userId)
-	if (BlockedList[userId] ~= nil and BlockedList[userId] == true) then
+	if (BlockedList[userId]) then
```
hockeychaos: what is the content of the value here? basically we are just checking if it isn't nil now, but is that enough?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript.rbxmx
```diff
@@ -320,11 +325,6 @@ LocalPlayer.CharacterAdded:connect(function(character)
 		ControlState:SwitchTo(lastControlState)
 	end
 end)
-
-LocalPlayer.CharacterRemoving:connect(function(character)
```
hockeychaos: if the character is removed, do we still show controls on the screen? It seems like we might want to keep this code as well as add it to characteradded

CoreScriptsRoot/Modules/PlayerDropDown.lua
garnold24: This was some leftover code from debugging that I forgot to remove. I was initially using that to check if the issue was coming from the web request.

CoreScriptsRoot/Modules/PlayerDropDown.lua
garnold24: BlockedList should now be a dictionary of user ids, structured like

key        value
userid    true

Since it only depends on the userId being a key, it doesn't matter what the value is.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript.rbxmx
garnold24: I'll do that.

PR #418, opened 2016-06-09
VR-39: Recenter View button and various cleanup/fixes

kyle-emmerich:


CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -957,17 +979,42 @@ local function Create3DSettingsIcon(topBarInstance, panel, menubar)
 
 		BackgroundTransparency = 1,
 
-		Image = "rbxasset://textures/ui/Menu/hamburger3D.png"
+		Image = "rbxasset://textures/ui/VR/hamburger.png"
 	}
 
+	local menuState = false
```
SolarCrane: This variable name isn't very descriptive of what it does, if it is used to keep track of whether the menu is open or not we should call it something like `menuOpened` or `menuOpenState`

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -957,17 +979,42 @@ local function Create3DSettingsIcon(topBarInstance, panel, menubar)
...
+	GuiService.MenuOpened:connect(function() menuState = true end)
+	GuiService.MenuClosed:connect(function() menuState = false end)
+
+	function menuItem:OnSetActive(active)
+		MenuModule:SetVisibility(active)
+		if not active then
```
SolarCrane: I am not sure what is going on here:
* How come we are waiting for menuState to go false after setting the menu item to not active? Is there a spawn or a wait in the `MenuModule:SetVisibility` and is the function calling `OnSetActive` need to be assured that the menu is closed?
* This wait loop is concerning, we could break the corescripts if it sits around waiting forever
* We usually liked to put Async at the end of functions like this to let the caller know that it could stall

Ultimately, can we remove this wait and rather use an event to let the called know when it is closed?

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -957,17 +979,42 @@ local function Create3DSettingsIcon(topBarInstance, panel, menubar)
...
+		isTransitioningVR = false
 	end)
 	
 	local settingsPanel = Panel3D.Get("SettingsMenu")
 	function settingsPanel:OnVisibilityChanged(visible)
-		menuItem:SetActive(visible)
+		menuItem:SetActive(visible, true)
+		if not visible then
+			while menuState do
+				game:GetService("RunService").RenderStepped:wait()
```
SolarCrane: Can we also replace this while wait with an event?

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: Hmm, the point was to make sure CloseAllExcept would stall until the animation completes. Perhaps it would be better to do GuiService.MenuClosed:wait(), now that I think of it. I'll change CloseAllExcept to CloseAllExceptAsync.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -957,17 +979,34 @@ local function Create3DSettingsIcon(topBarInstance, panel, menubar)
...
 		BackgroundTransparency = 1,
 
-		Image = "rbxasset://textures/ui/Menu/hamburger3D.png"
+		Image = "rbxasset://textures/ui/VR/hamburger.png"
 	}
 
+	function menuItem:OnSetActive(active)
+		MenuModule:SetVisibility(active)
+		if not active and MenuModule:GetVisibility() then
+			GuiService.MenuClosed:wait()
```
SolarCrane: Again, is it necessary to wait for the close event in `OnSetActive`

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -957,17 +979,34 @@ local function Create3DSettingsIcon(topBarInstance, panel, menubar)
...
+		menubar:CloseAllExceptAsync(menuItem, true)
+		menuItem:SetActive(not menuItem:IsActive())
+		isTransitioningVR = false
 	end)
 	
 	local settingsPanel = Panel3D.Get("SettingsMenu")
 	function settingsPanel:OnVisibilityChanged(visible)
-		menuItem:SetActive(visible)
+		if not visible and MenuModule:GetVisibility() then
+			GuiService.MenuClosed:wait()
```
SolarCrane: Here too

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -415,15 +418,34 @@ local function Create3DMenuItem()
 	}
 
 	local isActive = false
+	local isModal = true
```
SolarCrane: I am uneasy about keeping track of isModal here, it feels like we are over-coupling the menubar icons with the panels

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -351,10 +352,12 @@ local function Create3DMenuBar(barAlignment, threeDPanel)
 		return totalWidth
 	end
 
-	function this:CloseAllExcept(item)
+	function this:CloseAllExceptAsync(item, ignoreModality)
 		for _, v in pairs(this:GetItems()) do
 			if v ~= item then
-				v:SetActive(false)
+				if ignoreModality or v:IsModal() then
```
SolarCrane: Just checking that if `ignoreModality` is false and `v:IsModal()` is also false we don't want to set that item to inactive?

kyle-emmerich: Correct. It will only be hidden if it's modal.

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: This is a separate concept from panel modal-ness. Should we perhaps name it "isExclusive" or something?

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: We could either yield here or add a version of MenuModule:SetVisibility() that yields. Bad things happen if you don't wait for the animation to complete. Hm.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -536,6 +534,13 @@ local function Create3DMenuItem()
 		self:SetHoverText(hoverTextLabel.Text)
 	end
 
+	local defaultClickBehavior = backgroundButton.MouseButton1Click:connect(function()
```
SolarCrane: I typically like to name these with `Connection` or `Conn` at the end of the variable name to signify that it is a connection

kyle-emmerich: Ah, I forgot to remove this.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1277,23 +1289,41 @@ local function Create3DUserGuiToggleIcon(topBarInstance, panel, menubar)
...
+		end
+	end
+
 	function userGuiPanel:OnVisibilityChanged(visible)
 		if visible then
 			local headLook = Panel3D.GetHeadLookXZ(true)
 			userGuiPanel.localCF = headLook * CFrame.Angles(math.rad(5), 0, 0) * CFrame.new(0, 0, 5)
 		end
-		menuItem:SetActive(visible)
+		menuItem:SetActive(visible, true)
```
SolarCrane: Is there a `SetActive` that takes 2 args? It looks like the initial definition only takes 1 arg?

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1277,23 +1289,41 @@ local function Create3DUserGuiToggleIcon(topBarInstance, panel, menubar)
...
-	userGuiPanel:ResizePixels(300, 125)
+	userGuiPanel:ResizeStuds(4, 4, 128)
 	userGuiPanel:SetVisible(false, false)
 
-	menuItem.MouseButton1Click:connect(function()
-		if InputService.VREnabled then
-			userGuiPanel:SetVisible(not userGuiPanel.isVisible, false)
+	local VRHub = require(GuiRoot.Modules.VR.VRHub)
+
+	VRHub.ModuleOpened.Event:connect(function(moduleName, isExclusive, shouldCloseNonExclusive, shouldKeepTopbarOpen)
```
SolarCrane: This seems like a lot of args for this event, and it is getting kind of confusing what they all do. If VRHub is keeping state of these things perhaps we can call some functions on the VRHub to properly process this event?

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1277,23 +1289,41 @@ local function Create3DUserGuiToggleIcon(topBarInstance, panel, menubar)
...
+		if shouldCloseNonExclusive and menuItem:IsActive() then
+			menuItem:SetActive(false)
+			userGuiPanel:SetVisible(false)
+
+			VRHub:FireModuleClosed(thisModuleName)
```
SolarCrane: Seems funny to fire other modules events, but it maybe ok here.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1451,13 +1656,19 @@ if nameAndHealthMenuItem and not isTenFootInterface then
 end
 
 if settingsIcon3D then
-	THREE_D_ITEM_ORDER[settingsIcon3D] = 1
+	ITEM_ORDER_3D[settingsIcon3D] = 1
 end
 if chatIcon3D then
-	THREE_D_ITEM_ORDER[chatIcon3D] = 2
+	ITEM_ORDER_3D[chatIcon3D] = 4
```
SolarCrane: Sort of odd these are out of order now, lets re-order them to be 1-5

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1588,26 +1799,53 @@ if userGuiIcon3D and vr3dGuis then
...
+			TopbarPanel3D:SetVisible(true)
+
+			modulesKeepingTopbarOpen[moduleName] = true
+			numModulesKeepingTopbarOpen = numModulesKeepingTopbarOpen + 1
+		end
+	end)
+	VRHub.ModuleClosed.Event:connect(function(moduleName)
+		if modulesKeepingTopbarOpen[moduleName] then
+			modulesKeepingTopbarOpen[moduleName] = nil
+			numModulesKeepingTopbarOpen = math.max(0, numModulesKeepingTopbarOpen - 1)
```
SolarCrane: This feels like it could get out of sync somehow. If we could somehow ask our modules if they are open / need to keep topbar open that would feel better. Plus if we did that we could get rid of the `shouldKeepTopbarOpen` variable

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -129,6 +141,40 @@ function Panel3D.SetModalPanel(panel)
...
+		local isOnGui = false
+
+		if lookX >= guiX and lookX <= guiX + guiWidth and
+		   lookY >= guiY and lookY <= guiY + guiHeight then
+		   	isOnGui = true
+		end
+
+		return worldIntersectPoint, localIntersectPoint, lookAtPixel, isOnGui
+	else
+		return nil, nil, false
```
SolarCrane: shouldn't the third return value here be a Vector2 or nil and false be the fourth return value?

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -129,6 +141,40 @@ function Panel3D.SetModalPanel(panel)
...
+	local planeNormal = planeCF.lookVector
+	local pointOnPlane = planeCF.p + (planeNormal * partThickness * 0.5)
+
+	local worldIntersectPoint = Util.RayPlaneIntersection(ray, planeNormal, pointOnPlane)
+	if worldIntersectPoint then
+		local parentGuiWidth, parentGuiHeight = parentGui.AbsoluteSize.X, parentGui.AbsoluteSize.Y
+		local localIntersectPoint = (planeCF:pointToObjectSpace(worldIntersectPoint) / currentHeadScale) * Vector3.new(-1, 1, 1) + Vector3.new(partWidth / 2, -partHeight / 2, 0)
+		local lookAtPixel = Vector2.new((localIntersectPoint.X / partWidth) * parentGuiWidth, (localIntersectPoint.Y / partHeight) * -parentGuiHeight)
+		
+		--fire mouse enter/leave events if necessary
```
SolarCrane: I thought we fixed the need to manually do enter and leave events if lua

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -628,32 +713,194 @@ end
...
+function Subpanel.new(parentPanel, guiElement)
+	local self = setmetatable({}, Subpanel_mt)
+	self.parentPanel = parentPanel
+	self.guiElement = guiElement
+	self.lastParent = guiElement.Parent
+	self.ancestryConn = nil
+	self.changedConn = nil
+
+	self.lookAtPixel = Vector2.new(0,0)
+	self.lookedAt = fals
```
SolarCrane: this line just feels like it is lacking somethin

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1403,10 +1433,183 @@ local function CreateStopRecordIcon()
...
+
+	return menuItem
+end
+------------------------
+
 local TopBar = CreateTopBar()
 local LeftMenubar = CreateMenuBar(BarAlignmentEnum.Left)
 local RightMenubar = CreateMenuBar(BarAlignmentEnum.Right)
-local ThreeDMenubar = Create3DMenuBar(BarAlignmentEnum.Left, Topbar3DPanel)
+local Menubar3D = CreateMenuBar3D(BarAlignmentEnum.Left, TopbarPanel3D)
```
SolarCrane: I know this isn't your change, but it seems like we are mixing names here; we should decide to call it Topbar or Menubar

CoreScriptsRoot/Modules/VR/Panel3D.lua
BinaryResolved: subtle

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: I thought the Menubar was what contained MenuItems and the topbar contains multiple menubars.

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: Hey, that works too.

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: We did, but this covers the entire panel. I'll have to check if we actually use it still.

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: Huh. You're right. I remember fixing this but it didn't make it into this commit. I don't see anything else missing though.

CoreScriptsRoot/Modules/VR/Panel3D.lua
kyle-emmerich: Of course. It's missing "local fals = false". 

On a more serious note, nil evaluates to false when coerced to boolean... so functionally speaking it isn't wrong...

kyle-emmerich:
I don't like Git

PR #424, opened 2016-06-13
CLI-10725 Buttons should be smaller in the UI for mobile

garnold24:
Made the mobile Jump and context action buttons the same size as the interior thumbstick button (35x35 pixels on small devices, 45,45 on larger ones) and then positioned them accordingly to have roughly the same position on the other side of the screen.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/ContextActionTouch.lua
+++ b/CoreScriptsRoot/CoreScripts/ContextActionTouch.lua
@@ -194,6 +194,9 @@ function createButton( actionName, functionInfoTable )
 
 	if buttonScreenGui and buttonScreenGui.Parent == nil then
 		buttonScreenGui.Parent = localPlayer.PlayerGui
+		if not buttonFrame.Parent then
+			buttonFrame.Parent = buttonScreenGui
+		end
 	end
 end
 
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/TouchJump.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/TouchJump.rbxmx
@@ -56,9 +56,9 @@ function TouchJump:Create(parentFrame)
 	JumpButton.ImageRectOffset = Vector2.new(176, 222)
 	JumpButton.ImageRectSize = Vector2.new(174, 174)
 	JumpButton.Size = UDim2.new(0, jumpButtonSize, 0, jumpButtonSize)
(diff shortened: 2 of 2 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/TouchJump.rbxmx
```diff
@@ -55,8 +56,8 @@ function TouchJump:Create(parentFrame)
 	JumpButton.ImageRectOffset = Vector2.new(176, 222)
 	JumpButton.ImageRectSize = Vector2.new(174, 174)
 	JumpButton.Size = UDim2.new(0, jumpButtonSize, 0, jumpButtonSize)
-	JumpButton.Position = isSmallScreen and UDim2.new(1, jumpButtonSize * -2.25, 1, -jumpButtonSize - 20) or
-		UDim2.new(1, jumpButtonSize * -2.75, 1, -jumpButtonSize - 120)
+	JumpButton.Position = isSmallScreen and UDim2.new(1, -origJumpButtonSize - 10, 1, -origJumpButtonSize) or
```
hockeychaos: has this been tested on phones (big and small phones) and tablets? Want to make sure the positioning seems ok across all devices

garnold24: I've tested it on my phone, an iPhone 4s with a tiny screen, and a Samsung tablet with a normal aspect ratio. I can try a few more to make sure it looks ok on more devices.

garnold24: It was positioned a little high on an iPad, so I adjusted it to be level with the thumbstick. I tried it out on a few more devices and it looks good on all.

CoreScriptsRoot/CoreScripts/ContextActionTouch.lua
```diff
@@ -194,6 +194,9 @@ function createButton( actionName, functionInfoTable )
 
 	if buttonScreenGui and buttonScreenGui.Parent == nil then
 		buttonScreenGui.Parent = localPlayer.PlayerGui
+		if not buttonFrame.Parent then
```
hockeychaos: are we going to fix this to follow playergui lifetime?

garnold24: I will look into this, but regardless; I think we should keep this fix because there's no downside to it and it still works as a solution to the issue if we can't figure out why it's getting parented to nil on respawn.

PR #431, opened 2016-06-20
CLIMOB-2466 Fixed other Movement Modes not working after spawning wit…

garnold24:
…h ClickToMove enabled

Because of how ClickToMove isn't a module of the control script, the necessary code for connecting the movement style changes wasn't running, due to it not getting a module. This fix makes it so that these events are called outside of that if statement. So far this change works fine on my mobile devices; but I'll look into the ClickToMove module and Camera Script to make sure there aren't any unforeseen issues.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript.rbxmx
@@ -155,9 +155,9 @@ ControlModules.Touch.GameSettingsChangedCon = nil
 function ControlModules.Touch:RefreshControlStyle()
 	if ControlModules.Touch.Current then
 		ControlModules.Touch.Current:Disable()
-		setJumpModule(false)
-		TouchJumpModule:Disable()
 	end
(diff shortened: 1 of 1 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript.rbxmx
```diff
@@ -153,10 +153,11 @@ ControlModules.Touch.LocalPlayerChangedCon = nil
 ControlModules.Touch.GameSettingsChangedCon = nil
 
 function ControlModules.Touch:RefreshControlStyle()
-	ControlModules.Touch.Current:Disable()
-	setJumpModule(false)
-	TouchJumpModule:Disable()
-	
+	if ControlModules.Touch.Current then
+		ControlModules.Touch.Current:Disable()
+		setJumpModule(false)
```
hockeychaos: do the jump module lines need to be in this if?

garnold24: No, they are best left out of the if statement. I overlooked that.

PR #437, opened 2016-06-25
Interns have blue text

Quenty:


kyle-emmerich:
Is this approved?

hockeychaos:
If we can get a roblox employee to own this group and manage membership I'm game to merge this in

kyle-emmerich:
Why not Ben? 😃

hockeychaos:
lol who is in charge of all of you? It should probably be that person

hockeychaos:
also I think if we want to do this we need a new intern chat color + icon in the player list

kyle-emmerich:
Mr. CodeWriter might be a good choice then, he seems to be the client team's Lord of the Flies.

How about blue?

talbotwhite:
I think red would be better.

RobloxCodeWriter:
Red is not a good color for messaging.  Red is the color for errors or
warning and we don't want every message looking like something's wrong.

garnold24:
I suggest making the text Silver or Bronze. It goes with the theme of admins having yellow/golden text.

BinaryResolved:
If that's gonna be the group used shouldn't we have more than 3 interns in the group?

kyle-emmerich:
If this goes in, we can get all the interns to join pretty quickly.

BinaryResolved:
If so then should past interns have the chat color and badge? Because there's a rank called 'Former Interns'.

einsteinK:
Anyone checked the current group used?
It has a rank "Intern" already:
https://www.roblox.com/groups/group.aspx?gid=1200769

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -270,22 +270,30 @@ do
...
-				adminCache[userId] = isAdmin
+	local function MakeIsInGroup(groupId, requiredRank)
+		assert(type(requiredRank) == "nil" or type(requiredRank) == "number", "requiredRank must be a number or nil")
+		
+		local inGroupCache = {}
+		return function(player)
+			if player and player.userId then
+				local userId = player.userId
+
+				if inGroupCache[userId] then
```
einsteinK: Eh, did you forget a ` == nil` or am I imagining things?
It seems as if this will **always** be *false*, resulting it never being set.

kyle-emmerich: Yeah this should have ==nil (and not just "not inGroupCache...")

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -270,22 +270,30 @@ do
...
+					local inGroup = false
+					pcall(function() -- Many things can error is the IsInGroup check
+						inGroup = (not requiredRank) and player:IsInGroup(groupId) or (player:GetRankInGroup(groupId) > requiredRank)
+					end)
+					inGroupCache[userId] = inGroup
+				end
+
+				return inGroupCache[userId]
 			end
-			return adminCache[userId]
```
matthewdean: Shouldn't this be return inGroupCache[userId]?

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -270,22 +270,30 @@ do
...
+		
+		local inGroupCache = {}
+		return function(player)
+			if player and player.userId then
+				local userId = player.userId
+
+				if inGroupCache[userId] == nil then
+					local inGroup = false
+					pcall(function() -- Many things can error is the IsInGroup check
+						inGroup = (not requiredRank) and player:IsInGroup(groupId) or (player:GetRankInGroup(groupId) > requiredRank)
```
matthewdean: As much as your psuedo-ternary operator coding is appreciated, I think it would be more straightforward just to do:

if requiredRank then
  inGroup = player:GetRankInGroup(groupId) > requiredRank
else
  inGroup = player:IsInGroup(groupId)
end

CoreScriptsRoot/Modules/Chat.lua
Quenty: You're looking at the old code.

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -270,22 +270,34 @@ do
...
-				pcall(function()
-					isAdmin = player:IsInGroup(1200769)
-				end)
-				adminCache[userId] = isAdmin
+	local function MakeIsInGroup(groupId, requiredRank)
+		assert(type(requiredRank) == "nil" or type(requiredRank) == "number", "requiredRank must be a number or nil")
+		
+		local inGroupCache = {}
+		return function(player)
+			if player and player.userId then
```
matthewdean: and player.UserId > 0

CoreScriptsRoot/Modules/Chat.lua
Quenty: Fixed!

spotco:
Just checking: is there a reason why this hasn't been accepted yet?

kyle-emmerich:
The code is kind of messy. We should have a cleanup round first. Also not sure if the group is owned by a full staff member yet, is it?

Quenty:
Cleanup, group ownership, and final approval are all here as soon as @matthewdean commits changes.

PR #438, opened 2016-06-28
VR-184: Voice to text button for VR keyboard

SolarCrane:


CoreScriptsRoot/Modules/VR/VirtualKeyboard.lua
```diff
@@ -186,6 +187,40 @@ local MINIMAL_KEYBOARD_LAYOUT_SYMBOLS = HttpService:JSONDecode([==[
...
+	    'ASR_STATUS_DATA_LOSS',
+	    -- last official google response
+
+	    -- Roblox statuses
+	    'ASR_STATUS_NOT_ENABLED',
+	    'ASR_STATUS_LOW_CONFIDENCE',
+	    'ASR_STATUS_INVALID_JSON'
+	};
+
+	for i, code in pairs(STATUS_CODES) do
```
jmargh: Think we should use ipairs or a for i = 1, style loop, I believe pairs does not guarantee order, even for array style tables.

PR #440, opened 2016-07-01
CLI-11547 Dev console mobile improvements

garnold24:
Added some quality of life fixes to the dev console on mobile devices.

- The minimum size of the console is now 400x200 pixels to make sure all
of the console is navigable on smaller mobile devices.
- Added Touch events to allow the console to be resized, moved, and
scrollable on mobile devices.

### DIFF
--- a/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
+++ b/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
@@ -273,11 +273,15 @@ end
 
 function Methods.ResetFrameDimensions(devConsole)
 	devConsole.Frame.Size = UDim2_new(0.5, 20, 0.5, 20);
-	devConsole.Frame.Position = UDim2_new(0.25, -10, 0.125, -10)
+	
+	local abSize = devConsole.Frame.AbsoluteSize
+	devConsole:SetFrameSize(abSize.x, abSize.y)
+	local newSize = devConsole.Frame.Size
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -273,11 +273,15 @@ end
 
 function Methods.ResetFrameDimensions(devConsole)
 	devConsole.Frame.Size = UDim2_new(0.5, 20, 0.5, 20);
-	devConsole.Frame.Position = UDim2_new(0.25, -10, 0.125, -10)
+	
+	local abSize = devConsole.Frame.AbsoluteSize
+	devConsole:SetFrameSize(abSize.x, abSize.y)
```
hockeychaos: I'm sure there is a reason for this, but it seems weird. Could you explain the reasoning here? maybe we should comment it

garnold24: This is a check to make sure that when the developer console first appears, it isn't smaller than what is defined as the minimum size. I tried to keep it so that it wouldn't mess with its original behavior. I could rewrite some of it to make it more elegant.

garnold24: I should also mention that I added code to SetFrameSize to check if it's too small.

PR #445, opened 2016-07-08
Started a new Utility module for all CoreScripts to use, and converte…

kyle-emmerich:
…d Panel3D to use it. (Also small fixes for Panel3D)

Fixes in Panel3D:
* Use new Utility module
* Make references to game.Players consistent (and use GetService to avoid bugs when devs rename services)
* Make SurfaceGuis and Parts created by the panel system Archivable=false just in case. This matters more for the Parts since they exist in the camera for now.


kyle-emmerich:
Let's not merge this yet, name collision problems with Modules.Utility

kyle-emmerich:
Going to wait on this!

PR #454, opened 2016-07-15
CLIXBOX-833 - Let devs open dev console on Xbox

jmargh:


jmargh:
This allows place/group owners to open the developer console on Xbox. Adjusted some selection issues with gamepad navigation, removed the old dev console script in favor of moving things to the module itself.

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -1823,6 +1828,19 @@ do
 		textBox.TextYAlignment = 'Top'
 		textBox.FontSize = 'Size18'
 		textBox.TextWrapped = true
+
+		-- override SelectionImageObject to better fit
+		if isTenFootInterface then
+			local selectionImage = Instance.new('ImageLabel')
```
SolarCrane: We should probably also name this `selectionImage`

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -2929,4 +2952,61 @@ do
...
+
+		if isTenFootInterface then
+			myDeveloperConsole.VisibleChanged:connect(onDevConsoleVisibilityChanged)
+		end
+	end
+
+	return myDeveloperConsole
+end
+
+function DevConsoleModuleTable:ToggleVisibility()
```
SolarCrane: It may be better in the long run to just expose the GetVisible and SetVisible functions. Since toggle only really works for this use case

CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
```diff
@@ -738,6 +738,41 @@ local function Initialize()
...
+		if game.CreatorType == Enum.CreatorType.Group then
+			spawn(function()
+				-- spawn since GetRankInGroup is async
+				local success, result = pcall(function()
+					return LocalPlayer:GetRankInGroup(game.CreatorId) == 255
+				end)
+				if success and result == true then
+					makeDevConsoleOption()
+				else
+					print("DeveloperConsoleModule: GetRankInGroup failed because", result)
```
SolarCrane: Will this print every time someone queries it and isn't the group owner even if the call is successful?

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -2198,6 +2198,16 @@ local function CreateChat()
 			pcall(function()
 				if sendingPlayer == PlayersService.LocalPlayer then
 					isDebugCommand = game:GetService("GuiService"):ShowStatsBasedOnInputString(chattedMessage)
+					
+					-- allows dev console to be opened on mobile
+					-- NOTE: Removed ToggleDevConsole bindable event, so engine no longer handles this
+					if string.lower(chattedMessage) == "/console" then
+						local devConsoleModule = require(RobloxGui.Modules.DeveloperConsoleModule)
```
SolarCrane: Might want to add a `if not NON_CORESCRIPT_MODE then` here to prevent the script from crashing if devs copy our corescripts

jmargh: I will add it to the first if check then since ShowStatsBasedOnInputString() is RobloxScript only.

PR #455, opened 2016-07-16
New Chat System for Code Review

JacobLocke:
Not currently in a state for deployment. This will be updated over time with fixes and improvements.
Started pull request now for code review.

Edit: It is actually now totally ready to be pulled in!


CoreScriptsRoot/Modules/ChatSelector.lua
```diff
@@ -0,0 +1,203 @@
+local USE_NEW_CHAT = false --and settings():GetFFlag("UseNewChat")
```
SolarCrane: It seems like the GetFFlag here was commented out for testing purposes, if so we should put it back in.

Also when using FFlag we should use this format:
local pcallSuccess, flagEnabled= pcall(function() return settings():GetFFlag("YourFlag") end)
local isItReallyOn = pcallSuccess and flagEnabled

JacobLocke: It was commented out because the flag doesn't exist yet. It's just there right now as a reminder to implement later. I guess a comment there also would have been good. I'll be sure to update it to that format later when the flag is created.

kyle-emmerich: If you use the format above you should be able to have the code in without the flag even existing, so no need to worry about updating it later.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,222 @@
...
+	return setmetatable(obj, metatable)
+end
+
+return module
+
+]]
+
+local generated = Instance.new("ModuleScript")
+generated.Name = "Generated"
+generated.Source = source
```
SolarCrane: I am not terribly fond of scripts creating scripts. Is there a way to avoid this or is it necessary to create a new script?

JacobLocke: These scripts are loaded as instance type of CoreScript, but they need to be ModuleScripts. There isn't a way to load ModuleScripts on the server side in corescripts right now, so I wrote this code to "solve" that problem. It works for now and it's not necessarily all that bad, but it could be changed if this functionality was natively added.

CoreScriptsRoot/Modules/ChatSelector.lua
```diff
@@ -0,0 +1,203 @@
...
+
+		return sig
+	end
+end
+
+
+local moduleApiTable = {}
+do
+	if (USE_NEW_CHAT) then
+		local StarterGui = game:GetService("StarterGui")
```
SolarCrane: Rather than having GetService for StarterGui in multiple spots lets just put all the GetServices at the top of the file

CoreScriptsRoot/ServerCoreScripts/ClientChat/NewChat.lua
```diff
@@ -0,0 +1,415 @@
...
+	moduleApiTable.ChatBarFocusChanged = Util.Signal()
+	moduleApiTable.VisibilityStateChanged = Util.Signal()
+	moduleApiTable.MessagesChanged = Util.Signal()
+
+
+	moduleApiTable.MessagePosted = Util.Signal()
+	moduleApiTable.CoreGuiChanged = Util.Signal()
+	
+	
+	moduleApiTable.eChatMakeSystemMessage = Util.Signal()
```
SolarCrane: Rather than prefixing `e` here, can we add an `Event` to the end of the variable names?

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatService.lua
```diff
@@ -0,0 +1,285 @@
...
+
+
+--////////////////////////////// Methods
+--//////////////////////////////////////
+function metatable:Dump()
+	return tostring(self)
+end
+
+function metatable:DoMessageFilter(speakerName, message)
+	for i, func in pairs(self.FilterMessageFunctions) do
```
SolarCrane: If we don't use `i` we should replace it with a `_`

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatService.lua
```diff
@@ -0,0 +1,285 @@
...
+		pcall(function()
+			local ret = func(speakerName, message)
+			assert(type(ret) == "string")
+			message = ret
+		end)
+	end
+	
+	message = self.FilterMessageFunctions["default_filter"](speakerName, message)
+	
+	if (game:GetService("RunService"):IsServer() and not game:GetService("RunService"):IsStudio()) then
```
SolarCrane: Feels like we should get the RunService once and put it in a variable instead of multiple times

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatService.lua
```diff
@@ -0,0 +1,285 @@
...
+	return message
+end
+
+function metatable:DoProcessCommands(speakerName, message, channel)
+	local processed = false
+	
+	processed = self.ProcessCommandsFunctions["default_commands"](speakerName, message)
+	if (processed) then return processed end
+	
+	for i, func in pairs(self.ProcessCommandsFunctions) do
```
SolarCrane: Same as above

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatService.lua
```diff
@@ -0,0 +1,285 @@
...
+		
+		if (processed) then break end
+	end
+	
+	return processed
+end
+
+function metatable:AddChannel(channelName)
+	if (self.ChatChannels[channelName:lower()]) then
+		error("Channel \"" .. channelName .. "\" already exists!")
```
SolarCrane: Long strings like this we should use `string.format` on

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatService.lua
```diff
@@ -0,0 +1,285 @@
...
+function metatable:RemoveChannel(channelName)
+	if (self.ChatChannels[channelName:lower()]) then
+		local n = self.ChatChannels[channelName:lower()].Name
+		
+		self.ChatChannels[channelName:lower()]:Destroy()
+		self.ChatChannels[channelName:lower()] = nil
+
+		spawn(function() self.eOnChannelRemoved:Fire(n) end)
+	else
+		warn("Channel \"" .. channelName .. "\" does not exist!")
```
SolarCrane: string.format these please

CoreScriptsRoot/ServerCoreScripts/ServerChat/SpeakerProxy.lua
```diff
@@ -0,0 +1,148 @@
...
+	
+	function obj:GetExtraData(key)
+		return obj.Target:GetExtraData(key)
+	end
+	
+	metatable.__metatable = "The metatable is locked"
+	metatable.__tostring = obj.Target.__tostring
+	
+	metatable.__index = function(tbl, index)
+		if (index == "Name" or
```
SolarCrane: This if statement is a bit excessive. Is there no way to automate the checks here? I feel like if we add a function then we would have to alter these lines way down here. If not we should at least have a look-up table so we don't have to cram a bunch of `or` here

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatServiceProxy.lua
```diff
@@ -0,0 +1,212 @@
...
+		
+		obj.Target:UnregisterProcessCommandsFunction(funcId)
+	end
+	
+	local metatable = getmetatable(proxy)
+	metatable.__metatable = "The metatable is locked"
+	metatable.__tostring = function () return tostring(obj.Target) end
+	
+	metatable.__index = function(tbl, index)
+		if (index == "ChatLogLength" or
```
SolarCrane: This is an awful lot of `or`s flying around here. Can we put this in a lookup table?

kyle-emmerich: Looks like you could do without this if statement entirely and just check if the result of the indexing was nil or not.

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -2756,4 +2756,3 @@ do
 end
 
 return moduleApiTable
-
```
spotco: Revert this whitespace change for style.

JacobLocke:
List of new client sided files that were changed in new commit:
All in directory "ServerCoreScripts\ClientChat"
ChannelsBar.lua
ChannelsTab.lua
ChatBar.lua
ChatChannel.lua
ChatScript.lua
ChatWindow.lua
ChatWindowInstaller.lua
MessageLabelCreator.lua
NewChat.lua
SpeakerDatabase.lua
TransparencyTweener.lua

Modules/ChatSelector.lua was also changed significantly.

Still not entirely ready for the pull request to go through, but significant progress was made in the redesign and was ready for reviewing.

CoreScriptsRoot/Modules/ChatSelector.lua
```diff
@@ -0,0 +1,184 @@
...
+	spawn(function()
+		local useNewChat = GetUseLuaFlag()
+		local useModuleScript = (useNewChat or FORCE_USE_NEW_CHAT) and RobloxGui.Modules.NewChat or RobloxGui.Modules.Chat
+		useModule = require(useModuleScript)
+
+		ConnectSignals(useModule, interface, "ChatBarFocusChanged")
+		ConnectSignals(useModule, interface, "VisibilityStateChanged")
+		ConnectSignals(useModule, interface, "MessagesChanged")
+
+		StarterGui:RegisterGetCore("UseNewLuaChat", function() return useNewChat end)
```
SolarCrane: Have an event?

CoreScriptsRoot/Modules/NewChat.lua
```diff
@@ -0,0 +1,236 @@
+local CoreGuiService = game:GetService("CoreGui")
+local RobloxGui = CoreGuiService:WaitForChild("RobloxGui")
+
+local StarterGui = game:GetService("StarterGui")
+local GuiService = game:GetService("GuiService")
+
+local Util = {}
```
SolarCrane: Move chat stuff into its own folder and makes a util module for these things.

CoreScriptsRoot/Modules/ChatSelector.lua
```diff
@@ -0,0 +1,162 @@
...
+			useModule:SetVisible(visible)
+		else
+			state.Visible = visible
+		end
+	end
+
+	function interface:FocusChatBar()
+		if (useModule) then
+			useModule:FocusChatBar()
+		else
```
SolarCrane: Can probably just delete this empty else block

CoreScriptsRoot/Modules/ChatSelector.lua
```diff
@@ -0,0 +1,162 @@
...
+			return useModule:GetMessageCount()
+		else
+			return 0
+		end
+	end
+
+	function interface:TopbarEnabledChanged(...)
+		if (useModule) then
+			return useModule:TopbarEnabledChanged(...)
+		else
```
SolarCrane: Delete empty else block here

CoreScriptsRoot/Modules/ChatSelector.lua
JacobLocke: I wanted to keep it in there like that to keep the same format that the other functions had and to blatantly say to not do anything in this situation.

CoreScriptsRoot/Modules/ChatSelector.lua
```diff
@@ -0,0 +1,162 @@
...
+StarterGui:RegisterGetCore("ChatWindowSize", NonFunc)
+StarterGui:RegisterSetCore("ChatBarDisabled", NonFunc)
+StarterGui:RegisterGetCore("ChatBarDisabled", NonFunc)
+
+
+local function ConnectSignals(useModule, interface, sigName)
+	useModule[sigName]:connect(function(...) interface[sigName]:fire(...) end)
+end
+
+local isConsole = GuiService:IsTenFootInterface() 		or false -- set true to force a true value
```
SolarCrane: Lets get rid of these crazy spaces and the `or false`. If we want to have a way to force it on we could put a flag variable at the top

CoreScriptsRoot/Modules/ChatSelector.lua
```diff
@@ -0,0 +1,162 @@
...
+	end
+
+	interface.ChatBarFocusChanged = Util.Signal()
+	interface.VisibilityStateChanged = Util.Signal()
+	interface.MessagesChanged = Util.Signal()
+end
+
+local stopCachingMakeSystemMessage = false
+local MakeSystemMessageCache = {}
+local function MakeSystemMessageCachingFunction(data)
```
SolarCrane: It seems like this function just queues up system messages until we are ready. We should probably rename it to something like `QueueSystemMessage` so that it is a little less confusing.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatWindow.lua
```diff
@@ -0,0 +1,305 @@
...
+--//////////////////////////////////////
+local function CreateGuiObject()
+	local BaseFrame = Instance.new("Frame")
+	BaseFrame.BackgroundTransparency = 1
+	BaseFrame.Size = UDim2.new(0.35, 0, 0.35, 0)
+
+	--// This spawns a new thread that waits approximately long enough 
+	--// until the BaseFrame has been parented to PlayerGui so it can 
+	--// actually check the size of it.
+	spawn(function()
```
SolarCrane: This is a potential race-condition, we should connect to a changed event here and update when the size changes to avoid timing issues.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatWindow.lua
```diff
@@ -0,0 +1,305 @@
...
+		wait()
+		if (BaseFrame.AbsoluteSize.X < 1600/3.75) then
+			BaseFrame.Size = UDim2.new(0, 1600/3.75, 0, 900/3.75)
+		end
+	end)
+
+	--// Chat Size18, 8 pixels on each end for white box in center, 
+	--// 8 pixels on each end for actual chatbot object
+	local chatBarYSize = 18 + 16 + 16
```
SolarCrane: Instead of hard-coding 18 here we should extract the height from the font size we are using (in-case we ever change the size or allow the user to customize their font size)

JacobLocke: This change is already under way in what I'm going to update soon. It also has dynamic font sizing based on a config table.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatWindow.lua
```diff
@@ -0,0 +1,305 @@
...
+	ChatChannelParentFrame.BackgroundTransparency = 0.6
+	ChatChannelParentFrame.BorderSizePixel = 0
+
+	local Players = game:GetService("Players")
+	if (not Players.ClassicChat and Players.BubbleChat) then
+		ChatBarParentFrame.Position = UDim2.new(0, 0, 0, 0)
+		ChannelsBarParentFrame.Visible = false
+		ChatChannelParentFrame.Visible = false
+
+		spawn(function()
```
SolarCrane: What is with the `wait()`s here? Can we also use an event here?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatWindow.lua
```diff
@@ -0,0 +1,305 @@
...
+function metatable:RemoveChannel(channelName)
+	if (not self:GetChannel(channelName))  then
+		error("Channel '" .. channelName .. "' does not exist!")
+	end
+	
+	local indexName = channelName:lower()
+	
+	if (self.Channels[indexName] == self:GetCurrentChannel()) then
+		if (indexName == "all") then
+			self:SwitchCurrentChannel(nil)
```
SolarCrane: In this case shouldn't we try to switch to any open chat channel?

CoreScriptsRoot/ServerCoreScripts/ServerChat/Speaker.lua
```diff
@@ -0,0 +1,187 @@
+local source = [[
+local module = {}
+
+--////////////////////////////// Details
+--//////////////////////////////////////
+local metatable = {}
+metatable.__ClassName = "Speaker"
+
+metatable.__tostring = function(tbl)
```
SolarCrane: Seems like we do this metatable business a lot, could we make a generic `MakeClass` function? Or something to that extent?

kyle-emmerich: I would personally opt for a simpler approach in the form of this kind of pattern:

local YourClass = {}
YourClass.__index = YourClass
function YourClass.new()
     local self = setmetatable({}, YourClass)
     self.member = 5
     return self
end

function YourClass:__tostring() ... end
function YourClass:SomeMethod() ... end

That way there's not really much room to wonder what it does and you get your metamethods and normal methods all in one place. Checking the type is just checking getmetatable(instance)==YourClass.

CoreScriptsRoot/Modules/NewChat.lua
```diff
@@ -0,0 +1,233 @@
...
+		StarterGui:RegisterSetCore("ChatMakeSystemMessage", function(data)
+			local event = FindIndexInCollectionWithType(communicationsConnections.SetCore, "ChatMakeSystemMessage", "BindableEvent")
+			if (event) then
+				event:Fire(data)
+			else
+				table.insert(MakeSystemMessageCache, data)
+			end
+		end)
+		
+		StarterGui:RegisterGetCore("ChatWindowPosition", function(data)
```
SolarCrane: All these functions seem very copy pasted, we should refactor them into the one function that we call in these functions.

CoreScriptsRoot/Modules/NewChat.lua
```diff
@@ -0,0 +1,233 @@
...
+		
+		moduleApiTable.ChatBarFocusChanged = Util.Signal()
+		moduleApiTable.VisibilityStateChanged = Util.Signal()
+		moduleApiTable.MessagesChanged = Util.Signal()
+
+		StarterGui.CoreGuiChangedSignal:connect(function(coreGuiType, enabled)
+			if (coreGuiType == Enum.CoreGuiType.All or coreGuiType == Enum.CoreGuiType.Chat) then
+				ChatWindowState.CoreGuiEnabled = enabled
+
+				local event = FindIndexInCollectionWithType(communicationsConnections.ChatWindow, "CoreGuiEnabled", "BindableEvent")
```
SolarCrane: This event dispatching code seems to be very similar, we should refactor it into its own function

CoreScriptsRoot/Modules/NewChat.lua
```diff
@@ -0,0 +1,233 @@
...
+			TopbarEnabled = true,
+			CoreGuiEnabled = true,
+		}
+
+		local communicationsConnections = {}
+		local eventConnections = {}
+
+		local MakeSystemMessageCache = {}
+
+		local function FindIndexInCollectionWithType(collection, indexName, type)
```
SolarCrane: This name is misleading, because you are finding the Value stored and not the index; `FindInCollectionByKeyAndType` would read a lot better

CoreScriptsRoot/Modules/NewChat.lua
```diff
@@ -0,0 +1,224 @@
...
+				local rVal = nil
+				if (func) then rVal = func:Invoke(data) end
+				return rVal
+			end)
+		end
+
+		StarterGui.CoreGuiChangedSignal:connect(function(coreGuiType, enabled)
+			if (coreGuiType == Enum.CoreGuiType.All or coreGuiType == Enum.CoreGuiType.Chat) then
+				ChatWindowState.CoreGuiEnabled = enabled
+				DispatchEvent("CoroeGuiEnabled", ChatWindowState.CoreGuiEnabled)
```
SolarCrane: Is `CoroeGuiEnabled` a typo?

CoreScriptsRoot/Modules/NewChat.lua
```diff
@@ -0,0 +1,224 @@
...
+				return func:Invoke(useWasFocused)
+			end
+			return false
+		end
+		
+		moduleApiTable.ChatBarFocusChanged = Util.Signal()
+		moduleApiTable.VisibilityStateChanged = Util.Signal()
+		moduleApiTable.MessagesChanged = Util.Signal()
+
+		local function DispatchEvent(eventName, ...)
```
SolarCrane: Why not use this in the above functions like `TopbarEnabledChanged` etc...

CoreScriptsRoot/Modules/NewChat.lua
JacobLocke: most likely

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,313 @@
...
+	LeaveConfirmationNotice.BackgroundTransparency = 1
+	LeaveConfirmationNotice.TextColor3 = Color3.new(1, 1, 1)
+	LeaveConfirmationNotice.TextStrokeTransparency = 0.75
+	LeaveConfirmationNotice.Text = "Leave channel <XX>?"
+	LeaveConfirmationNotice.Font = Enum.Font.SourceSansBold
+	LeaveConfirmationNotice.FontSize = Enum.FontSize.Size18
+
+	local outPos = LeaveConfirmationFrame.Position
+	LeaveConfirmationButtonYes.MouseButton1Click:connect(function()
+		print("Leave channel")
```
SolarCrane: Might want to comment these out for shipping to prod.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,313 @@
...
+	tab.GuiObject.Parent = self.ScrollerFrame
+	self.ChannelTabs[channelName:lower()] = tab
+
+	self.NumTabs = self.NumTabs + 1
+	self:OrganizeChannelTabs()
+
+	self.BackgroundTweener:RegisterTweenObjectProperty(tab.BackgroundTweener, "Transparency")
+	self.TextTweener:RegisterTweenObjectProperty(tab.TextTweener, "Transparency")
+
+	--tab.NameTag.MouseButton2Click:connect(function()
```
SolarCrane: Is this commented out code something for the future, or can we remove it?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,313 @@
...
+	if (tab) then
+		tab:UpdateMessagePostedInChannel()
+	else
+		warn("ChannelsTab '" .. channelName .. "' does not exist!")
+	end
+end
+		
+function methods:AddChannelTab(channelName)
+	if (self:GetChannelTab(channelName)) then
+		error("Channel tab '" .. channelName .. "'already exists!")
```
SolarCrane: Is it okay for this function to crash? Other calling scripts are protected from potential execution stoppage?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsTab.lua
```diff
@@ -0,0 +1,251 @@
...
+	obj.ChannelName = channelName
+	obj.UnreadMessageCount = 0
+	obj.Active = false
+
+	obj.BackgroundTweener = moduleTransparencyTweener.new()
+	obj.TextTweener = moduleTransparencyTweener.new()
+
+	obj.GuiObject.Name = "Frame_" .. obj.ChannelName
+
+	local maxNameLenght = 12
```
SolarCrane: Might be a little typo here with `length`

JacobLocke: I am aware of this text clipping for long names issue as in it doesn't work perfectly. It's a coming soon (tm) thing for second round improvements.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
JacobLocke: This function is called from another function that is called from a RemoteEvent. The other function has a similar check and error in it and if either of them crash then nothing is created, the one thread is stopped, and an error message is outputted to the console. It doesn't stop execution of the rest of the system however.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatBar.lua
```diff
@@ -0,0 +1,246 @@
...
+	TextLabel.BackgroundTransparency = 1
+	TextLabel.Size = TextBox.Size
+	TextLabel.Position = TextBox.Position
+	TextLabel.FontSize = TextBox.FontSize
+	TextLabel.Font = TextBox.Font
+	TextLabel.TextColor3 = TextBox.TextColor3
+	TextLabel.TextStrokeTransparency = TextBox.TextStrokeTransparency
+	TextLabel.TextXAlignment = TextBox.TextXAlignment
+	TextLabel.TextYAlignment = TextBox.TextYAlignment
+	TextLabel.Text = "This value was un-initialized lololololololololol"
```
SolarCrane: This is an interesting text string

JacobLocke: The idea was that a function would be called later to set the default text based on if it was mobile or PC. This way you could make it say "click to chat" or "touch to chat".

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatBar.lua
```diff
@@ -0,0 +1,246 @@
...
+
+	obj.TweenPixelsPerSecond = 500
+	obj.TargetYSize = 0
+
+	obj.BackgroundTweener = moduleTransparencyTweener.new()
+	obj.TextTweener = moduleTransparencyTweener.new()
+
+	ClassMaker.MakeClass("ChatBar", obj)
+
+	obj:SetTextLabelText("To chat click here or press \"/\" key")
```
SolarCrane: You can use single quotes or the [[ ]] string creation to avoid complicated escapes

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,322 @@
...
+	tab.GuiObject.Parent = self.ScrollerFrame
+	self.ChannelTabs[channelName:lower()] = tab
+
+	self.NumTabs = self.NumTabs + 1
+	self:OrganizeChannelTabs()
+
+	self.BackgroundTweener:RegisterTweenObjectProperty(tab.BackgroundTweener, "Transparency")
+	self.TextTweener:RegisterTweenObjectProperty(tab.TextTweener, "Transparency")
+
+	--// Although this feature is pretty much ready, it needs some UI design still.
```
kyle-emmerich: Any update on that, or can it be shipped as is and updated later?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,322 @@
...
+	ClassMaker.MakeClass("ChannelsBar", obj)
+
+	obj:CreateTweeners()
+
+	PageLeftButton.MouseButton1Click:connect(function() obj:ScrollChannelsFrame(-1) end)
+	PageRightButton.MouseButton1Click:connect(function() obj:ScrollChannelsFrame(1) end)
+
+	--// Have to wait to tween until it's properly parented to PlayerGui
+	spawn(function()
+		wait()
```
kyle-emmerich: Is there an event you could wait for rather than just a shotgun kind of approach? Maybe you could do some AncestryChanged:wait() here.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatBar.lua
```diff
@@ -0,0 +1,246 @@
...
+function methods:CalculateSize()
+	local lastPos = self.GuiObject.Size
+	self.GuiObject.Size = UDim2.new(1, 0, 0, 1000)
+
+	local fontSize = tonumber(self.TextBox.FontSize.Name:match("%d+"))
+	local bounds = self.TextBox.TextBounds.Y
+
+	self.GuiObject.Size = lastPos
+
+	local newTargetYSize = bounds - fontSize --+ 4
```
kyle-emmerich: Could be confusing in the future. Remove or add a note?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
JacobLocke: It's super basic and wouldn't look well with the current look of the UI and I'm not a UI designer so I don't know how to make it look good on my own.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,739 @@
...
+end
+
+--// Event for focusing the chat bar when player presses "/".
+local ChatBarUISConnection = UserInputService.InputBegan:connect(function(input)
+	if (input.KeyCode == Enum.KeyCode.Slash) then
+		DoChatBarFocus()
+	end
+end)
+
+--ChatBarUISConnection:disconnect()
```
kyle-emmerich: Is this supposed to be here?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,739 @@
...
+end)
+
+EventFolder.OnChannelLeft.OnClientEvent:connect(function(channel)
+	ChatWindow:RemoveChannel(channel)
+
+	DoFadeInFromNewInformation()
+end)
+
+EventFolder.OnMuted.OnClientEvent:connect(function(channel)
+	-- handle
```
kyle-emmerich: did you handle it

CoreScriptsRoot/ServerCoreScripts/ClientChat/TransparencyTweener.lua
```diff
@@ -0,0 +1,196 @@
...
+			addStr =  addStr .. " [" .. propertyName .. "=" .. string.sub(tostring(baseValue), 1, 4) .. "]"
+		end
+
+		str = str .. addStr
+	end
+
+	return str
+end
+
+function metatable:OutputTest()
```
kyle-emmerich: I actually really like that you left this in. It'll be nice for whoever needs to maintain this in the future if you're not around to help. Maybe we should do this sort of thing more.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatChannel.lua
```diff
@@ -0,0 +1,155 @@
...
+function methods:PositionMessageLabelInWindow(messageObject)
+	local baseFrame = messageObject.BaseFrame
+	local baseMessage = messageObject.BaseMessage
+
+	baseFrame.Parent = self.Scroller
+	baseFrame.Position = UDim2.new(0, 0, 0, self.Scroller.CanvasSize.Y.Offset)
+	
+	--// This looks stupid, but it's actually necessary.
+	--// TextBounds wont be calculated correctly unless it has enough space.
+	baseFrame.Size = UDim2.new(1, 0, 0, 1000)
```
SolarCrane: We already have the textbounds calculator that was copied from the original chat scripts somewhere in this commit

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
JacobLocke: It ended up being unused. Before we got art from the art team and the chat window UI looked bad, this feature was used to take away the chat bar when a player was muted in a channel. What was found out was that it was very inconvenient when a player couldn't chat out commands like "/join channel" when they were muted in every channel they were in. When recreating the window UI script after getting the art, this part was added in but nothing was done with it.

I'll update the comment to be more clear about this in case in the future something will want to be done with it.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatChannel.lua
kyle-emmerich: Is this intended to eventually be dev-space code? If not, could use TextService.

JacobLocke: I don't think this particular situation is a big deal. It's only one extra line rather than calling a function from another module to do the calculation. Plus, the comments make it clear what is happening.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+		if (numChildrenRemaining < 1) then
+			FoundAllEventsEvent:Fire()
+		end
+	end)
+
+	FoundAllEventsEvent.Event:wait() -- I love how this turned out
+	con:disconnect()
+
+	FoundAllEventsEvent:Destroy()
+end
```
SolarCrane: This all seems extraneous, we could just have an array of children and iterate over it calling WaitForChild() on each of them. IIRC WaitForChild only yields if the child doesn't exist yet.

JacobLocke: It has to be this way because there are weird situations like if there are children with the same name but different class types and/or not all children are added yet. For example if someone had a RemoteFunction with the same name as one of the RemoteEvents we were waiting for in an already existing EventFolder. WaitForChild would find that RemoteFunction first and then not be aware of the RemoteEvent that would  be added later and then the script would break.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+		DoBackgroundFadeIn()
+		DoTextFadeIn()
+	else
+		DoBackgroundFadeIn()
+	end
+end
+
+
+local last = 0
+spawn(function()
```
SolarCrane: The preferred way to do RenderStepped code like this is rather to call `BindToRenderStep` on the RunService

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+			fadedChanged.Event:wait()
+
+		end
+
+	end
+end)
+
+
+local function UpdateMousePosition(mousePos)
+	if (not moduleApiTable.Visible or not moduleApiTable.IsCoreGuiEnabled or not moduleApiTable.TopbarEnabled) then return end
```
SolarCrane: demorgan's law simplifies this to:
`not (moduleApiTable.Visible and moduleApiTable.IsCoreGuiEnabled and moduleApiTable.TopbarEnabled)`

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
JacobLocke: There's also other Event:waits() in this loop so this loop doesn't have to run every single render frame. Doing BindToRenderStepped would force it to run every single render frame which is not necessarily what is wanted at every moment.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+end
+
+UserInputService.InputChanged:connect(function(inputObject)
+	if (inputObject.UserInputType == Enum.UserInputType.MouseMovement) then
+		local mousePos = Vector2.new(inputObject.Position.X, inputObject.Position.Y)
+		UpdateMousePosition(mousePos)
+	end
+end)
+
+UserInputService.TouchTap:connect(function(tapPos, gameProcessedEvent)
```
SolarCrane: Not sure if this is intentional, but I believe this a high level touch event and only fire on quick tap gestures rather than any touch movement.

JacobLocke: It was intentional so that players have to tap in the chat window UI space to activate it. They could potentially want to rotate the camera through that space or something and not want to activate it.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+	local last = mouseIsInWindow
+
+	UpdateMousePosition(tapPos[1])
+	if (not mouseIsInWindow  and last ~= mouseIsInWindow)  then
+		DoBackgroundFadeOut()
+	end
+end)
+
+UpdateFadingForMouseState(true)
```
SolarCrane: Comment or address?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+	-- handle
+end)
+
+EventFolder.OnSpeakerExtraDataUpdated.OnClientEvent:connect(function(speakerName, data)
+	local speaker = SpeakerDatabase:GetSpeaker(speakerName)
+	if (not speaker) then
+		speaker = SpeakerDatabase:AddSpeaker(speakerName)
+	end
+	
+	for k, v in pairs(data) do
```
SolarCrane: What is k and v here? It would be great to have some descriptive names here

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+		ChatWindow:SwitchCurrentChannel(channel)
+	end
+end)
+
+
+local reparentingLock = false
+
+--// Do not remove on death behavior
+LocalPlayer.CharacterRemoving:connect(function()
+	if (true or reparentingLock) then return end
```
SolarCrane: What is the purpose of this? We are always going to return right away?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+
+	GuiParent.Parent = nil
+	LocalPlayer.CharacterAdded:wait()
+	GuiParent.Parent = PlayerGui
+
+end)
+
+local function connectGuiParent(GuiParent)
```
SolarCrane: Not sure what is going on here, what is the destroyGuard for?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+			connectGuiParent(GuiParent)
+		end
+	end)
+end
+
+connectGuiParent(GuiParent)
+
+--// Always on top behavior that relies on parenting order of ScreenGuis
+--// This would end up really bad if something else tried to do the exact same thing however.
+PlayerGui.ChildAdded:connect(function(child)
```
SolarCrane: This function also always returns right away?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+
+
+EventFolder.OnMainChannelSet.OnClientEvent:connect(function(channel)
+	if (ChatWindow:GetChannel(channel)) then
+		ChatWindow:SwitchCurrentChannel(channel)
+	end
+end)
+
+
+local reparentingLock = false
```
SolarCrane: This variable is always false it seems as it is, do we need a reparentingLock? Users can already disable the core chat

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,735 @@
...
+
+
+	function moduleApiTable:SpecialKeyPressed(key, modifiers)
+		if (key == Enum.SpecialKey.ChatHotkey) then
+			DoChatBarFocus()
+		end
+	end
+end
+
+spawn(function() wait() moduleApiTable:SetVisible(false) moduleApiTable:SetVisible(true) end)
```
SolarCrane: spawn already does a wait. Do we need to wait for this function invocation?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
JacobLocke: It was supposed to be used, but it wasn't implemented. In the fix I pushed it is now used.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatBar.lua
SolarCrane: Seems a tad bit unprofessional though eh

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatChannel.lua
SolarCrane: Just one line here and there can add up, especially if you want to change things later

kyle-emmerich: Any problem can be solved with another layer of indirection ;)

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
SolarCrane: I think the bug in that case is that there are RemoteFunctions inside the Events folder, especially with a name that is already in use.

JacobLocke: The thing is that it's exposed to developers so they can add in any random things they want to. There also may be a pre-existing folder with the same name we want to use and we can't just destroy the old one and create our own, but we can add a few events in for this system to use.

CoreScriptsRoot/Modules/ChatSelector.lua
spotco: No existing pattern (or existing, event would require a new RegisterCore functionality.
Simplified the way it is.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatWindow.lua
spotco: Addressing this. Will move any open channel.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatChannel.lua
spotco: This will be addressed.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
spotco: This one should be kept as is.
The key point is that the client type checks the objects here, and properly waits for server replication of creation of new object of correct type/name if it doesn't.

CoreScriptsRoot/ServerCoreScripts/ServerChat/README.md
```diff
@@ -0,0 +1,106 @@
...
+	ChatChannel GetChannel(string channelName)
+
+	Speaker AddSpeaker(string speakerName)
+	void RemoveSpeaker(string speakerName)
+	Speaker GetSpeaker(string speakerName)
+
+	string[] GetChannelList()
+	string[] GetAutoJoinChannelList()
+
+	void AddWordFilter(string expression)
```
hockeychaos: we should remove these for now

CoreScriptsRoot/ServerCoreScripts/ServerChat/README.md
```diff
@@ -0,0 +1,106 @@
...
+	string Name (read-only)
+	string WelcomeMessage
+
+	bool Joinable
+	bool Leavable
+	bool AutoJoin
+	bool Private
+
+#### Methods
+	void AddWordFilter(string expression)
```
hockeychaos: we should remove these for now

CoreScriptsRoot/ServerCoreScripts/ServerChat/README.md
JacobLocke: oh right. Totally forgot

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,763 @@
...
+	end
+end)
+
+moduleApiTable.ChatMakeSystemMessageEvent:connect(function(valueTable)
+	if (valueTable["Text"] and type(valueTable["Text"]) == "string") then
+		local channel = "All"
+		local channelObj = ChatWindow:GetChannel(channel)
+
+		if (not channelObj) then
+			wait(0.25)
```
hockeychaos: why do we arbitrarily wait for a 1/4th of a second?

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatServiceRunner.lua
```diff
@@ -0,0 +1,275 @@
...
+	modules.ChildAdded:connect(function(child)
+		pcall(TryRunModule, child)
+	end)
+	
+	for i, module in pairs(modules:GetChildren()) do
+		pcall(TryRunModule, module)
+	end
+end
+
+wait()
```
hockeychaos: why is there a wait here

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatServiceRunner.lua
```diff
@@ -0,0 +1,275 @@
...
+
+didInit = true
+
+wait()
```
hockeychaos: why is there a wait here

CoreScriptsRoot/ServerStarterScript.lua
```diff
@@ -54,3 +54,32 @@ local function setDialogInUse(player, dialog, value)
...
+	while loop_continue do
+		local success, retVal = pcall(function()
+			return game.IsSFFlagsLoaded
+		end)
+		if not success then
+			loop_continue = false
+		elseif retVal then
+			loop_continue = false
+		else
+			wait(0.1)
```
hockeychaos: @spotco this is why I wanted an event

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatServiceRunner.lua
spotco: These can be removed. Thanks for catching it.

CoreScriptsRoot/Modules/ChatUtil.lua
```diff
@@ -0,0 +1,34 @@
+local Util = {}
+do
+	function Util.Signal()
```
hockeychaos: doesn't this signal library already exist somewhere else?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,329 @@
...
+end
+
+function methods:ScrollChannelsFrame(dir)
+	if (self.ScrollChannelsFrameLock) then return end
+	self.ScrollChannelsFrameLock = true
+
+	local newPageNum = self.CurPageNum + dir
+	if (newPageNum < 0) then
+		newPageNum = 0
+	elseif (newPageNum > 0 and newPageNum + 4 > self.NumTabs) then
```
hockeychaos: whats with the number 4? why do we want to do something different if 4 tabs? If this is necessary we should define at top of script, instead of using magic number everywhere

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,329 @@
...
+
+function methods:FadeInText(duration)
+	self.TextTweener:Tween(duration, 0)
+end
+
+function methods:CreateTweeners()
+	self.BackgroundTweener:CancelTween()
+	self.TextTweener:CancelTween()
+
+	self.BackgroundTweener = moduleTransparencyTweener.new()
```
hockeychaos: do we need to remake tweens when we just cancelled them?

CoreScriptsRoot/Modules/ChatUtil.lua
JacobLocke: A similar version exists somewhere, but there are differences.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -0,0 +1,329 @@
...
+end
+
+--///////////////////////// Constructors
+--//////////////////////////////////////
+ClassMaker.RegisterClassType("ChannelsBar", methods)
+
+function module.new()
+	local obj = {}
+
+	local BaseFrame, ScrollerFrame, PageLeftButton, PageRightButton, LeaveConfirmationFrame, LeaveConfirmationNotice = CreateGuiObject()
```
hockeychaos: CreateGuiObject naming is confusing because we make a lot of gui objects. Seems like the function constructs the ChannelsBar gui in total. CreateChannelsBarGui() ?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatChannel.lua
```diff
@@ -0,0 +1,163 @@
+local source = [[
+local module = {}
+--////////////////////////////// Include
+--//////////////////////////////////////
+local modulesFolder = script.Parent
+local moduleTransparencyTweener = require(modulesFolder:WaitForChild("TransparencyTweener"))
```
hockeychaos: is it worth it having an entire module devoted to just transparency tweening? I think @SolarCrane has made a few tweening libraries that are pretty generic

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatChannel.lua
```diff
@@ -0,0 +1,163 @@
...
+	local scrollBarBottomPosition = (self.Scroller.CanvasSize.Y.Offset - self.Scroller.AbsoluteSize.Y)
+	local reposition = (self.Scroller.CanvasPosition.Y >= scrollBarBottomPosition)
+
+	local add = UDim2.new(0, 0, 0, baseFrame.Size.Y.Offset)
+	self.Scroller.CanvasSize = self.Scroller.CanvasSize + add
+
+	if (reposition) then
+		self.Scroller.CanvasPosition = Vector2.new(0, math.max(0, self.Scroller.CanvasSize.Y.Offset - self.Scroller.AbsoluteSize.Y))
+	else
+		local displayNewMessagesIfNotFullyScrolledDown = false
```
hockeychaos: this is always false, lets just pass false here. If needed, add a comment for clarity

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
```diff
@@ -0,0 +1,763 @@
...
+
+
+
+
+
+
+--////////////////////////////////////////////////////////////////////////////////////////////
+--///////////// Code to talk to topbar and maintain set/get core backwards compatibility stuff
+--////////////////////////////////////////////////////////////////////////////////////////////
+local Util = {}
```
hockeychaos: this is the same util signal I saw above, please move this into one implementation. What is missing from the other signal util that is not possible to use here?

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatSettings.lua
```diff
@@ -0,0 +1,44 @@
+local source = [[
+-- This module is for initial settings that are not intended to be changed during runtime.
+
+local module = {}
+
+module.WindowDraggable = false
+module.WindowResizable = false -- soon (tee emm)
```
hockeychaos: I would prefer not to ship something that won't work, unless we plan on fixing this very soon

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
JacobLocke: TransparencyTweener is a data structure for tweening a group of objects as a single unit. In this method CreateTweeners() we first cancel any potential pre-existing tweens, and then recreate the tweener object so we can get an empty set of group data.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatMain.lua
JacobLocke: This one runs in developer space while the other one ran in core scripts.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatSettings.lua
JacobLocke: It's not a feature that absolutely needs to be completed for the chat system to function as well as the old one. The setting is just there as a placeholder for now. It's like #1 on the new features to-do list for afterwards.

CoreScriptsRoot/ServerCoreScripts/ClientChat/ChatChannel.lua
JacobLocke: It's for tweening groups of objects as a single unit, not for generic tweening of single objects.

PR #461, opened 2016-07-20
Don't check for gamepad until game loads

hockeychaos:


CoreScriptsRoot/CoreScripts/NotificationScript2.lua
```diff
@@ -835,6 +835,8 @@ if Platform == Enum.Platform.XBoxOne then
 		if controllerStateManager then
 			controllerStateManager:Initialize()
 
+			game.Loaded:wait()
+
```
jmargh: Can we do check for game:IsLoaded() before we connect to this?

hockeychaos: oh yeah duh

PR #462, opened 2016-07-21
CLIMOB-3088 Mobile Controls break when opening the settings while moving

garnold24:
The movement and jump controls for mobile would break whenever the settings menu opened while interacting with mobile controls. This fix goes through all the buttons and adds an event listener that stops movement when the menu opens.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/DPad.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/DPad.rbxmx
@@ -12,6 +12,7 @@
 --]]
 
 local Players = game:GetService('Players')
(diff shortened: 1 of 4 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/DPad.rbxmx
```diff
@@ -12,6 +12,7 @@
 --]]
 
 local Players = game:GetService('Players')
+local GUIS = game:GetService('GuiService')
```
hockeychaos: Could we name this GuiService? The other services are referenced in a way such that the name is basically the same

garnold24: Pretty sure I got them all.

PR #466, opened 2016-07-25
Reverted changing chat container from TextButton to Frame

garnold24:
This was still causing VR issues, and the best to go about fixing it seemed to just revert this code.

### DIFF
--- a/CoreScriptsRoot/Modules/Chat.lua
+++ b/CoreScriptsRoot/Modules/Chat.lua
@@ -1911,7 +1911,7 @@ local function CreateChatWindowWidget(settings)
 
 	local function CreateChatWindow()
 		-- This really shouldn't be a button, but it is currently needed for VR.
-		local container = Util.Create (InputService.VREnabled and 'TextButton' or 'Frame')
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -1911,7 +1911,7 @@ local function CreateChatWindowWidget(settings)
 
 	local function CreateChatWindow()
 		-- This really shouldn't be a button, but it is currently needed for VR.
-		local container = Util.Create (InputService.VREnabled and 'TextButton' or 'Frame')
+		local container = Util.Create 'TextButton'
```
hockeychaos: from the description I thought this would be a frame?

garnold24: Although it really should be a frame, For VR it's needed to be a button for the MouseButton1Click event. From my understanding, the reason the check didn't work is because VREnabled doesn't get set immediately when the game starts. So it would have to initialize as a frame, and then when VR does get enabled, change the entire hierarchy of the chat ui to make the base container a TextButton.

PR #470, opened 2016-07-27
CLIXBOX-842 - Allow users to open gamer cards from player list on Xbox

jmargh:


CoreScriptsRoot/Modules/PlayerlistModule.lua
```diff
@@ -781,7 +782,27 @@ function popupHidden()
...
+			local platformId = platformService:GetPlatformId(rbxUid)
+			if platformId and #platformId > 0 then
+				platformService:PopupProfileUI(Enum.UserInputType.Gamepad1, platformId)
+			end
+		end
+	end)
+end
+
 local function onEntryFrameSelected(selectedFrame, selectedPlayer)
+	if isTenFootInterface then
```
SolarCrane: It seems a bit odd here to check if isTenfootInterface is true and then later check PlatformService for the PlatformId. Is it possible to unify the two checks into checking the exact same thing?

CoreScriptsRoot/Modules/PlayerlistModule.lua
```diff
@@ -781,7 +782,27 @@ function popupHidden()
 end
 playerDropDown.HiddenSignal:connect(popupHidden)
 
+local function openPlatformProfileUI(rbxUid)
+	if not rbxUid or rbxUid < 1 then return end
+	pcall(function()
+		local platformService = game:GetService('PlatformService')
+		if platformService then
+			local platformId = platformService:GetPlatformId(rbxUid)
```
SolarCrane: What exactly is `platformId`? The name suggests it should be an int but it appears to be an array?

jmargh: GetPlatformId() returns a string, as the XDK API uses Platform::String^ for the xuid value. In addition, the PopupProfileUI() calls into an XDK function that expects Platform::String^.

Length is checked because the XDK function for opening the gamer card throws on empty string. The c++ function does account for this case, but I did an extra check here.

CoreScriptsRoot/Modules/PlayerlistModule.lua
jmargh: I'm a little confused on this. Do you mean we either only check isTenfootInterace or for PlatformService?

CoreScriptsRoot/Modules/PlayerlistModule.lua
```diff
@@ -1263,12 +1284,10 @@ local function createPlayerEntry(player, isTopStat)
 	local containerFrame, entryFrame = createEntryFrame(name, PlayerEntrySizeY, isTopStat)
 	entryFrame.Active = true
 
-	if not isTenFootInterface then
-		local function localEntrySelected()
-			onEntryFrameSelected(containerFrame, player)
-		end
-		entryFrame.MouseButton1Click:connect(localEntrySelected)
+	local function localEntrySelected()
```
SolarCrane: `localEntrySelected` isn't a very descriptive name for me. Also, if it is a callback to an event then we should start the name with `on`

CoreScriptsRoot/Modules/PlayerlistModule.lua
jmargh: Address by not checking for PlatformService in openPlatformProfuleUI()

CoreScriptsRoot/Modules/PlayerlistModule.lua
jmargh: Removed string length check to make less confusing. platformService:GetPlatformId() already checks string length.

CoreScriptsRoot/Modules/PlayerlistModule.lua
jmargh: removed localEntrySelected func and connect anonymous function to MouseButton1Click.

CoreScriptsRoot/Modules/PlayerlistModule.lua
SolarCrane: Okay, it seems like we could check `PlatformId ~= ""` if we want to check for that

CoreScriptsRoot/Modules/PlayerlistModule.lua
SolarCrane: What I was trying to get at was that we check `isTenFootInterface` and return early, but the function it calls may do nothing depending on whether the platform supports this behavior. Is this the desired behavior?

PR #471, opened 2016-07-28
VR-214: Revise Backpack based on Feedback

SolarCrane:


CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -36,18 +36,19 @@ local ARROW_IMAGE_CLOSE = 'rbxasset://textures/ui/Backpack_Close.png'
 local ARROW_SIZE = UDim2.new(0, 14, 0, 9)
 local ARROW_HOTKEY = Enum.KeyCode.Backquote.Value --TODO: Hookup '~' too?
 local ARROW_HOTKEY_STRING = '`'
+local ARROW_HOVER_COLOR = Color3.new(0,162/255,1)
 
 local HOTBAR_SLOTS_FULL = 10
+local HOTBAR_SLOTS_VR = 6
```
hockeychaos: should we rename the other SLOTS to be more descriptive? full/vr/mini seems like a weird split

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -58,6 +59,21 @@ local SEARCH_BACKGROUND_FADE = 0.15
 
 local DOUBLE_CLICK_TIME = 0.5
 
+local ZERO_KEY_VALUE = Enum.KeyCode.Zero.Value
```
hockeychaos: why not just use Enum.KeyCode.Zero.Value?

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -173,6 +189,55 @@ local function AdjustHotbarFrames()
...
+	InventoryFrame.Size = UDim2.new(0, HotbarFrame.Size.X.Offset, 0, (HotbarFrame.Size.Y.Offset * NumberOfInventoryRows) + INVENTORY_HEADER_SIZE + (IsVR and 2*INVENTORY_ARROWS_BUFFER_VR or 0))
+	InventoryFrame.Position = UDim2.new(0.5, -InventoryFrame.Size.X.Offset / 2, 1, HotbarFrame.Position.Y.Offset - InventoryFrame.Size.Y.Offset)
+
+	ScrollingFrame.Size = UDim2.new(1, ScrollingFrame.ScrollBarThickness + 1, 1, -INVENTORY_HEADER_SIZE - (IsVR and 2*INVENTORY_ARROWS_BUFFER_VR or 0))
+	ScrollingFrame.Position = UDim2.new(0, 0, 0, INVENTORY_HEADER_SIZE + (IsVR and INVENTORY_ARROWS_BUFFER_VR or 0))
+	AdjustHotbarFrames()
+	AdjustInventoryFrames()
+end
+
+local function SetNumberOfHotbarSlots(numSlots)
```
hockeychaos: this for addition/removal of vr headset right now?

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1066,13 +1141,15 @@ function enableGamepadInventoryControl()
 	ContextActionService:BindCoreAction("RBXBackpackHasGamepadFocus", noOpFunc, false, Enum.UserInputType.Gamepad1)
 	ContextActionService:BindCoreAction("RBXCloseInventory", goBackOneLevel, false, Enum.KeyCode.ButtonB, Enum.KeyCode.ButtonStart)
 
+	-- ContextActionService:BindCoreAction("RBXBackpackInventoryGamepadScroll", function() print("wut") end, false, Enum.UserInputType.ButtonL2, Enum.UserInputType.ButtonR2)
```
hockeychaos: why is this here?

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1172,90 +1267,102 @@ end
 
 -- Up arrow to open the inventory
 OpenInventoryButton = NewGui('ImageButton', 'OpenInventory')
-OpenInventoryButton.Size = UDim2.new(0, 44, 0, 40)
-OpenInventoryButton.Position = UDim2.new(0.5, -22, 1, -120)
-OpenInventoryButton.Image = "rbxasset://textures/ui/Backpack/ScrollUpArrow.png";
-OpenInventoryButton.MouseButton1Click:connect(function()
-	BackpackScript.OpenClose()
-end)
+do
```
hockeychaos: why do end? just to separate the code logically?

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1172,90 +1267,102 @@ end
...
 LeftBumperButton.Size = UDim2.new(0, 40, 0, 40)
 LeftBumperButton.Position = UDim2.new(0, -LeftBumperButton.Size.X.Offset, 0.5, -LeftBumperButton.Size.Y.Offset/2)
--- LeftBumperButton.Image = 'rbxasset://Textures/Sword128.png'
+LeftBumperButton.Image = "rbxasset://textures/ui/Backpack/ScrollUpArrow.png";
 
 RightBumperButton = NewGui('ImageLabel', 'RightBumper')
 RightBumperButton.Size = UDim2.new(0, 40, 0, 40)
 RightBumperButton.Position = UDim2.new(1, 0, 0.5, -RightBumperButton.Size.Y.Offset/2)
--- RightBumperButton.Image = 'rbxasset://Textures/Sword128.png'
+RightBumperButton.Image = "rbxasset://textures/ui/Backpack/ScrollUpArrow.png";
```
hockeychaos: is this probably not supposed to be the same image as LeftBumperButton

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1442,23 +1549,22 @@ do -- Search stuff
...
+		if input.UserInputType == Enum.UserInputType.Gamepad1 and not UserInputService.VREnabled then
 			searchFrame.Visible = false
 		else
 			searchFrame.Visible = true
 		end
 	end
-	local uis = game:GetService("UserInputService")
-	uis.InputBegan:connect(detectGamepad)
-	uis.InputChanged:connect(detectGamepad)
+	UserInputService.InputBegan:connect(detectGamepad)
```
hockeychaos: Might be better off using UserInputService.LastInputTypeChanged for this

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -97,6 +107,8 @@ local OpenInventoryButton = nil
 local CloseInventoryButton = nil
 local InventoryFrame = nil
 local ScrollingFrame = nil
+local ScrollUpInventoryButton = nil
+local ScrollDownInventoryButton = nil
```
jmargh: Outside of the scope of this task, but leaving comment for future reference. Should probably fix these forward declares and create the gui objects at the start of the script. Made it a bit confusing since they do get initialized in the main scope, but functions before they are initialized use them. Wasn't sure at first if they needed to be nil checked or not.

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1656,6 +1767,17 @@ local function OnVREnabled(prop)
 			end
 		end)
 
+		-- Turn off dragging when in VR
+		for _, slot in pairs(Slots) do
+			slot:SetClickability(false)
+		end
+
+		-- Reset the constants!
```
jmargh: What happens if VR is disabled? Is that even possible? [To get here UserInputService.VREanbled needs to be true.](https://github.com/ROBLOX/Core-Scripts/pull/471/files#diff-73dcf66ad2dbe31e15c6c4c51de4d67bL1578), so IsVR will always be true at this point, and the constants will not change.

CoreScriptsRoot/Modules/BackpackScript.lua
SolarCrane: We already have to support the mid-game enabling of the VR headset because `UserInputService.VREnabled` is not true when the game launches for a considerable duration.
I think it is a good idea to make it ready to disable VR headset in the scenario we will need to switch between the two.

CoreScriptsRoot/Modules/BackpackScript.lua
SolarCrane: wut, ok I shall deleets

CoreScriptsRoot/Modules/BackpackScript.lua
SolarCrane: Partly for organization of logic, but also because I am creating a local variable `openInventoryButtonSelectionObject` in this block that we shouldn't be referencing later.

CoreScriptsRoot/Modules/BackpackScript.lua
SolarCrane: We don't have art for these yet so they are actually not being shown. Eventually we will get new images for both left and right.

CoreScriptsRoot/Modules/BackpackScript.lua
SolarCrane: I will update that; I am guessing that when this was written we didn't have this API.

CoreScriptsRoot/Modules/BackpackScript.lua
SolarCrane: Right, I will address this by doing some updates that unlink the backpack from the VR interface

CoreScriptsRoot/Modules/BackpackScript.lua
SolarCrane: It is a very interesting design, however, as you mentioned I thought it would be best to conform with the existing design. If we want to change it we probably want an entire overhaul later.

CoreScriptsRoot/Modules/BackpackScript.lua
SolarCrane: I suppose the original author of this code thought it would be a worthy optimization!

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1441,24 +1559,23 @@ do -- Search stuff
...
 			searchFrame.Visible = false
 		else
 			searchFrame.Visible = true
 		end
 	end
-	local uis = game:GetService("UserInputService")
-	uis.InputBegan:connect(detectGamepad)
-	uis.InputChanged:connect(detectGamepad)
+	UserInputService.LastInputTypeChanged:connect(detectGamepad)
+	UserInputService.LastInputTypeChanged:connect(detectGamepad)
```
jmargh: Connecting to `LastInputTypeChanged` twice.

PR #472, opened 2016-07-28
Fix bug that breaks subpanels

kyle-emmerich:
Should note that this bug is harmless for non-VR although it might occasionally spam errors

### DIFF
--- a/CoreScriptsRoot/CoreScripts/Topbar.lua
+++ b/CoreScriptsRoot/CoreScripts/Topbar.lua
@@ -922,8 +922,8 @@ local function CreateLeaderstatsMenuItem()
 			end
 		end)
 
-	topbarEnabledChangedEvent.Event:connect(function()
-		PlayerlistModule.TopbarEnabledChanged(topbarEnabled)
+	topbarEnabledChangedEvent.Event:connect(function(enabled)
+		PlayerlistModule.TopbarEnabledChanged(enabled and not InputService.VREnabled) --We don't show the playerlist at all in VR
 	end)
 
 	this:SetColumns(PlayerlistModule.GetStats())
@@ -1191,8 +1191,8 @@ local function CreateChatIcon()
 		end
 	end
 
-	topbarEnabledChangedEvent.Event:connect(function()
-		ChatModule:TopbarEnabledChanged(topbarEnabled)
+	topbarEnabledChangedEvent.Event:connect(function(enabled)
+		ChatModule:TopbarEnabledChanged(enabled)
 	end)
 
 	chatIconButton.MouseButton1Click:connect(function()
@@ -1435,8 +1435,8 @@ local function CreateBackpackIcon()
 		BackpackModule:OpenClose()
 	end
 
-	topbarEnabledChangedEvent.Event:connect(function()
-		BackpackModule:TopbarEnabledChanged(topbarEnabled)
+	topbarEnabledChangedEvent.Event:connect(function(enabled)
+		BackpackModule:TopbarEnabledChanged(enabled)
 	end)
 
 	backpackIconButton.MouseButton1Click:connect(function()
@@ -1785,15 +1785,7 @@ end
 
 local function lookMenuEnabledChanged()
 	if InputService.VREnabled then
-		if lookMenuEnabled then
-			TopbarPanel3D:SetVisible(true)
-			--The look menu is very much tied to the backpack, so for now it will be responsible
-			--for showing/hiding it.
-			Panel3D.Get("Backpack"):SetVisible(StarterGui:GetCoreGuiEnabled(Enum.CoreGuiType.Backpack))
-		else
-			TopbarPanel3D:SetVisible(false)
-			Panel3D.Get("Backpack"):SetVisible(false)
-		end
+		TopbarPanel3D:SetVisible(lookMenuEnabled)
 	end
 end
 
@@ -1858,9 +1850,11 @@ local function topBarEnabledChanged()
 		lookMenuEnabled = topbarEnabled
 		topbarEnabled = false
 		lookMenuEnabledChanged()
+	else
+		lookMenuEnabled = false
 	end
 
-	topbarEnabledChangedEvent:Fire(topbarEnabled)
+	topbarEnabledChangedEvent:Fire(topbarEnabled or lookMenuEnabled)
 	TopBar:UpdateBackgroundTransparency()
 	for _, enumItem in pairs(Enum.CoreGuiType:GetEnumItems()) do
 		-- The All enum will be false if any of the coreguis are false
--- a/CoreScriptsRoot/CoreScripts/Topbar.lua
+++ b/CoreScriptsRoot/CoreScripts/Topbar.lua
@@ -693,7 +693,7 @@ local function CreateUsernameHealthMenuItem()
 	--------------
 
 	local function AnimateHurtOverlay()
-		if hurtOverlay and not VREnabled then
+		if hurtOverlay and not InputService.VREnabled then
 			local newSize = UDim2.new(20, 0, 20, 0)
 			local newPos = UDim2.new(-10, 0, -10, 0)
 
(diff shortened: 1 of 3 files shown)


CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -201,7 +202,7 @@ local function CreateTopBar()
 	end
 
 	function this:UpdateBackgroundTransparency()
-		if settingsActive and not VREnabled then
```
SolarCrane: If we are moving away from `VREnabled` variable here, we should delete all instances of it and get rid of the variable.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1782,12 +1783,24 @@ if userGuiIcon3D and vr3dGuis then
 	end
 end
 
+local function lookMenuEnabledChanged()
+	if InputService.VREnabled then
+		if lookMenuEnabled then
+			TopbarPanel3D:SetVisible(true)
+			--The look menu is very much tied to the backpack, so for now it will be responsible
+			--for showing/hiding it.
+			Panel3D.Get("Backpack"):SetVisible(StarterGui:GetCoreGuiEnabled(Enum.CoreGuiType.Backpack))
```
SolarCrane: Backpack is also calling the SetVisible functions on its panel. Is it possible to run this logic in Backpack module so that they don't fight with each other?

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1782,12 +1783,24 @@ if userGuiIcon3D and vr3dGuis then
...
+		else
+			TopbarPanel3D:SetVisible(false)
+			Panel3D.Get("Backpack"):SetVisible(false)
```
SolarCrane: Same as above.

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: Hm. It doesn't look like they'd fight, although it is kind of weird that panel:SetVisible happens here but the Backpack itself only toggles the guis.

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -1029,6 +1044,9 @@ local function onWorkspaceChanged(prop)
 			cameraChangedConnection:disconnect()
 		end
 		cameraChangedConnection = workspace.CurrentCamera.Changed:connect(onCameraChanged)
+
+		--Make sure the new camera gets used properly
+		onVREnabled("VREnabled")
```
SolarCrane: This is a bit misleading of a function invocation. We aren't actually enabling VR here or changing the state of VREnabled at all, rather we are trying refresh what is happening in that function. I would rather the onVREnabled function listen to an event that the CurrentCamera has changed, and stop listening when VREnabled is set to false.

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -41,6 +41,10 @@ local cursor = Utility:Create "ImageLabel" {
 	BackgroundTransparency = 1,
 	ZIndex = 10
 }
+local partFolder = Utility:Create "Folder" {
+	Name = "VR",
```
SolarCrane: This name is short, perhaps we should make it longer to avoid accidental collisions with developer named children (uhg, we have to muck up the devs workspace)

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1228,10 +1228,15 @@ end
...
 	-- Check for enabling/disabling the whole thing
 	if coreGuiType == Enum.CoreGuiType.Backpack or coreGuiType == Enum.CoreGuiType.All then
-		enabled = enabled and (TopbarEnabled or UserInputService.VREnabled)
+		enabled = enabled and TopbarEnabled 
 		WholeThingEnabled = enabled
 		MainFrame.Visible = enabled
 
+		if UserInputService.VREnabled then
+			local Panel3D = require(RobloxGui.Modules.VR.Panel3D)
+			Panel3D.Get("Backpack"):SetVisible(enabled)
```
SolarCrane: We set BackpackPanel to invisible when we open a module in VR:
` 			VRModuleOpenedConn = VRHub.ModuleOpened.Event:connect(function(moduleName)`
`				local openedModule = VRHub:GetModule(moduleName)`
`				if openedModule ~= BackpackScript and openedModule.VRIsExclusive then`
`					BackpackPanel:SetVisible(false)`

I think we should keep track of the variables controlling the visibility of Backpack and whenever we SetVisible it we should consult all those variables.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1879,4 +1890,7 @@ end
 
 -- Hook-up coregui changing
 StarterGui.CoreGuiChangedSignal:connect(OnCoreGuiChanged)
-topBarEnabledChanged()
+if not defeatableTopbar then
```
SolarCrane: Could this go in an else case of the if check up above?

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1840,7 +1846,15 @@ if gameOptions and not isTenFootInterface then
 end
 
 local function topBarEnabledChanged()
-	topbarEnabledChangedEvent:Fire(topbarEnabled)
+	if InputService.VREnabled then
+		lookMenuEnabled = topbarEnabled
+		topbarEnabled = false
```
SolarCrane: Seems a little dicey to me to change this property away from what the dev set it.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1782,12 +1783,16 @@ if userGuiIcon3D and vr3dGuis then
 	end
 end
 
+local function lookMenuEnabledChanged()
+	if InputService.VREnabled then
+		TopbarPanel3D:SetVisible(lookMenuEnabled)
```
SolarCrane: If we enable the look menu we probably want to change the SetGUIInsetBounds back to 0, 0, 0, 0

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: It should be okay if we think of it as transferring the state to lookMenuEnabled. I'll add preliminary support for toggling VR just so we don't forget to transfer back to topbarEnabled in the future.

CoreScriptsRoot/CoreScripts/Topbar.lua
kyle-emmerich: For reference, you'll find this on line 1857 of Topbar.lua

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1223,15 +1223,33 @@ end
 --| End Gamepad Functions |--
 -----------------------------
 
+local function EvaluateBackpackPanelVisibility(enabled)
+	if not enabled then
```
SolarCrane: Feel likes we should also check StarterGui:GetCoreGuiEnabled(Enum.CoreGuiType.Backpack)

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1223,15 +1223,36 @@ end
...
+		return false
+	end
+	if not TopbarEnabled then
+		return false
+	end
+	if not UserInputService.VREnabled then
+		return false
+	end
+
+	return true
```
SolarCrane: Isn't this whole function the same as:
` return enabled and StarterGui:GetCoreGuiEnabled(Enum.CoreGuiType.Backpack) and TopbarEnabled and UserInputService.VREnabled`

kyle-emmerich: well...

PR #475, opened 2016-08-01
Addressing feedback for VR-214

SolarCrane:
* Support X-button to swap a tool from hotbar<-->inventory
* Keep Search X button fully hidden when it cannot be pressed
* Keep Backpack tray depth the same when opening extended view
* Stop controller auto-select of first tool on open inventory
* Fix inventory arrow fade becoming desynced from Look Menu fade

### DIFF
--- a/CoreScriptsRoot/Modules/BackpackScript.lua
+++ b/CoreScriptsRoot/Modules/BackpackScript.lua
@@ -180,6 +180,10 @@ local function isInventoryEmpty()
 	return true
 end
 
+local function UseGazeSelection()
+	return UserInputService.VREnabled
+end
+
 local function AdjustHotbarFrames()
 	local inventoryOpen = InventoryFrame.Visible -- (Show all)
 	local visualTotal = (inventoryOpen) and NumberOfHotbarSlots or FullHotbarSlots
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -180,6 +180,10 @@ local function isInventoryEmpty()
 	return true
 end
 
+local function UseGazeSelection()
```
hockeychaos: I assuming this may change with daydream vs vive/etc right?

CoreScriptsRoot/Modules/BackpackScript.lua
```diff
@@ -1608,9 +1626,9 @@ do -- Make the Inventory expand/collapse arrow (unless TopBar)
 			disableGamepadInventoryControl()
 		end
 
-		if InventoryFrame.Visible and GamepadEnabled then
+		if InventoryFrame.Visible then
```
hockeychaos: why would we bind this if we don't have a gamepad? is this for if you connect one later?

SolarCrane: This is part of it, there is also an issue with GamepadEnabled being set to true due to funny reporting of the GamepadConnected events/functions. At any rate, it is listening to ButtonX (only exists on the gamepads)

PR #477, opened 2016-08-03
CLISTUDIO-10128: If no server is created in start server/player, player sees spam messages about WaitForChild

Quenty:
It appears this also contains changes from a separate pull request. Looking into it.


kyle-emmerich:
You'll probably just want to revert back to the commit before the one you want, stash the changes, and make a new branch, then apply the stashed changes on that branch.

CoreScriptsRoot/Modules/PlayerPermissionsModule.lua
```diff
@@ -0,0 +1,41 @@
+local PlayerPermissionsModule = {}
+
+local function HasRankInGroupFunctionFactory(groupId, requiredRank)
+	local hasRankCache = {}
```
hockeychaos: why don't we rely on the GetRankInGroup call to cache for us?

CoreScriptsRoot/Modules/PlayerlistModule.lua
```diff
@@ -263,20 +266,15 @@ local function getFollowerStatusIcon(followerStatus)
...
-	end)
-	if not success then
-		print("PlayerListScript2: getAdminIcon() failed because", result)
-		return nil
-	end
+	if CUSTOM_ICONS[userIdStr] then return nil end
 	--
-	if result then
+	
+	if PlayerPermissionsModule.IsPlayerAdminAsync(player) then
```
hockeychaos: seems a bit weird to not use an actual function call here

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -2215,7 +2233,7 @@ local function CreateChat()
 
 	function this:OnPlayerAdded(newPlayer)
 		if newPlayer then
-			spawn(function() Util.IsPlayerAdminAsync(newPlayer) end)
+			assert(coroutine.resume(coroutine.create(function() PlayerPermissionsModule.IsPlayerAdminAsync(newPlayer) end)))
```
hockeychaos: why did we switch from spawn?

CoreScriptsRoot/Modules/Chat.lua
```diff
@@ -274,22 +275,34 @@ do
...
-		local userId = player and player.userId
-		if userId then
-			if adminCache[userId] == nil then
-				local isAdmin = false
-				-- Many things can error is the IsInGroup check
-				pcall(function()
-					isAdmin = player:IsInGroup(1200769)
-				end)
-				adminCache[userId] = isAdmin
+	local function MakeIsInGroup(groupId, requiredRank)
```
hockeychaos: this is really similar to the player permissions module function, can we use that instead?

JetCrusherTorpedo:
Why not use `math.huge` for the timeout instead of `86400`?

CoreScriptsRoot/Modules/Chat.lua
Quenty: This was from a previous merge, it's literally an older version of the code, should be cleaned up once I do what Kyle said to do.

Quenty:
Waiting math.huge on ROBLOX returns instantly. 
`> wait(math.huge) print("Hi")`

Quenty:
Ok. Closing this pull request in favor of another one without the unnecessary changes. (Thanks Kyle!)

Issue #468, opened 2016-07-26
Infinite yield possible on 'RobloxReplicatedStorage:WaitForChild("SetDialogInUse")' 


JetCrusherTorpedo:
22:19:10.068 - Infinite yield possible on 'RobloxReplicatedStorage:WaitForChild("SetDialogInUse")'
22:19:10.070 - Stack Begin
22:19:10.070 - Script 'CoreGui.RobloxGui.CoreScripts/MainBotChatScript2', Line 71
22:19:10.070 - Stack End

PhoenixShay:
this is a problem with WaitForChild, not the corescripts themselves

someone at ROBLOX probably tried to make some sort of loop prevention on WaitForChild but never finished it

kyle-emmerich:
No, this is intended behavior. We knew that this sort of thing would show up going into this. You can safely ignore it.

What's happening is that the script uses WaitForChild to yield itself until SetDialogInUse is added to RobloxReplicatedStorage. This usage accounts for a very tiny proportion of what you'd usually use WFC for, so we provided that warning to make the normal usage easier.

PR #478, opened 2016-08-03
Removed unneeded warning message

JetCrusherTorpedo:
Fix issue #468

### DIFF
--- a/CoreScriptsRoot/CoreScripts/MainBotChatScript2.lua
+++ b/CoreScriptsRoot/CoreScripts/MainBotChatScript2.lua
@@ -68,7 +68,7 @@ local conversationTimedOut =        "Chat ended because you didn't reply"
 local conversationTimedOutSize = 350
 
 local RobloxReplicatedStorage = game:GetService('RobloxReplicatedStorage')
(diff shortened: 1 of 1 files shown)


JetCrusherTorpedo:
Note: using `math.huge` as the timeout does work. Open a new baseplate, type `print(Workspace:WaitForChild("Part",math.huge).Position)` in the command prompt. Wait as long as you like, then insert a part.

hockeychaos:
for consistency can you use the 24 hour wait convention? instead of math.huge?

JetCrusherTorpedo:
@hockeychaos Done

SolarCrane:
This feels like we are just running around the issue:
We implemented the potential infinite yield detection feature to identify possible bugs, so to say wait almost forever undermines what we set out to do.
Therefore, I think it would be better to hook this into a .ChildAdded event style and let the script fully execute.

JetCrusherTorpedo:
@SolarCrane How's that?

SolarCrane:
These updates seem a bit drastic, I think we should just rollback a couple commits to what @hockeychaos  recommended.

JetCrusherTorpedo:
@SolarCrane OK, done.

JetCrusherTorpedo:
@SolarCrane @hockeychaos Can this be merged?

PR #490, opened 2016-08-05
VR-178 Settings menu background in VR

kyle-emmerich:


CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -236,6 +236,18 @@ local function CreateSettingsHub()
 			ZIndex = SETTINGS_BASE_ZINDEX,
 			Parent = this.ClippingShield
 		};
+		this.VRShield = utility:Create("Frame") {
+			Name = "VRBackground",
+			Parent = this.Shield,
+
+			BackgroundColor3 = Color3.new(0.2, 0.2, 0.2),
```
SolarCrane: Are we able to use SETTINGS_SHIELD_COLOR for this shield as well? As well as SETTINGS_SHIELD_TRANSPARENCY or are they different values?
If they are different we could create some const for them at the top too.

CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -963,6 +975,7 @@ local function CreateSettingsHub()
 		
 		this.ClippingShield.Parent = panel:GetGUI()
 		this.Shield.Parent.ClipsDescendants = false
+		this.VRShield.Visible = true
```
SolarCrane: Should we also handle the case of turning off VR for future proofing? That could be in a separate commit though.

kyle-emmerich: I'll go ahead and add it since we're already here.

CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -984,15 +998,34 @@ local function CreateSettingsHub()
...
 	local UISChanged;
 	local function OnVREnabled(prop)
-		if prop == "VREnabled" and UserInputService.VREnabled then
-			VREnabled = true
-			enableVR()
-			if UISChanged then
-				UISChanged:disconnect()
-				UISChanged = nil
+		if prop == "VREnabled" then
+			VREnabled = UserInputService.VREnabled
```
SolarCrane: It seemed like you removed references to this variable, assuming you removed all of them we should remove this variable as well

kyle-emmerich:
Making a note because I just really confused myself: I removed an obsolete VREnabled variable from both of these pull requests. This one was in SettingsHub.lua while the other was in Topbar.lua.

PR #493, opened 2016-08-11
VR-232 Revise Notifications menu based on feedback

kyle-emmerich:


CoreScriptsRoot/Modules/VR/NotificationHub.lua
```diff
@@ -425,7 +446,9 @@ end
 NotificationGroup.new("Friends",        "Friends",  1)
 NotificationGroup.new("BadgeAwards",    "Badges",   2)
 NotificationGroup.new("PlayerPoints",   "Points",   3)
-NotificationGroup.new("Developer",      "Other",    4)
+
+--Don't want this right now.
+--NotificationGroup.new("Developer",      "Other",    4)
```
SolarCrane: For this kind of thing what I did in VR keyboard (for voice chat) was create a function at the top `IsDeveloperGroupEnabled()` function and just return false for now. Then when we want it we can turn make it true at a later point.

CoreScriptsRoot/Modules/VR/NotificationHub.lua
```diff
@@ -668,6 +697,13 @@ do
 		end
 	end)
 
+	local menuCloseShortcutBindName = game:GetService("HttpService"):GenerateGUID()
```
SolarCrane: Since these are core actions we do not need to generate random GUIDs for their names. It may be easier in some debugging scenario to have a legitimate name for this binding.

CoreScriptsRoot/Modules/VR/NotificationHub.lua
```diff
@@ -655,9 +688,13 @@ do
 
 	local SendNotificationInfoEvent = RobloxGui:WaitForChild("SendNotificationInfo")
 	SendNotificationInfoEvent.Event:connect(function(notificationInfo)
-		local group = notificationsGroups[notificationInfo.GroupName or "Developer"]
+		local group = notificationsGroups[notificationInfo.GroupName or false] --avoid error by nil index
```
SolarCrane: This line seems like it could be more elegantly written:
`local group = notificationInfo.GroupName and notificationsGroups[notificationInfo.GroupName]`

PR #494, opened 2016-08-11
Recenter is more usable and can be canceled via the following actions:

kyle-emmerich:
- Opening another module (settings, chat, notifications, etc)
 - Pressing the B button
 - Pressing the ESC button (will open settings)
 - Activating the recenter button again

### DIFF
--- a/CoreScriptsRoot/Modules/VR/Recenter.lua
+++ b/CoreScriptsRoot/Modules/VR/Recenter.lua
@@ -1,9 +1,12 @@
 local InputService = game:GetService("UserInputService")
+local ContextActionService = game:GetService("ContextActionService")
 local CoreGui = game:GetService("CoreGui")
 local RobloxGui = CoreGui:WaitForChild("RobloxGui")
 local Panel3D = require(RobloxGui.Modules.VR.Panel3D)
 local VRHub = require(RobloxGui.Modules.VR.VRHub)
 local Util = require(RobloxGui.Modules.Settings.Utility)
+
+local cancelShortcutName = game:GetService("HttpService"):GenerateGUID()
 
 local RecenterModule = {}
 RecenterModule.ModuleName = "Recenter"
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/VR/Recenter.lua
```diff
@@ -72,6 +78,12 @@ function RecenterModule:SetVisible(visible)
 			isCountingDown = true
 			countdownPanel:SetVisible(true)
 
+			ContextActionService:BindCoreAction(cancelShortcutName, function(actionName, inputState, inputObj)
+				if inputState == Enum.UserInputState.Begin then
+					cancelCountdown()
+				end
+			end, false, Enum.KeyCode.ButtonB)
+
```
SolarCrane: Is it possible to spawn off the count down logic? I don't think waiting in a `SetVisible` function is great because it yields other threads for something that doesn't seem like a yielding call.

CoreScriptsRoot/Modules/VR/Recenter.lua
```diff
@@ -1,10 +1,13 @@
 local InputService = game:GetService("UserInputService")
+local ContextActionService = game:GetService("ContextActionService")
 local CoreGui = game:GetService("CoreGui")
 local RobloxGui = CoreGui:WaitForChild("RobloxGui")
 local Panel3D = require(RobloxGui.Modules.VR.Panel3D)
 local VRHub = require(RobloxGui.Modules.VR.VRHub)
 local Util = require(RobloxGui.Modules.Settings.Utility)
 
+local cancelShortcutName = game:GetService("HttpService"):GenerateGUID()
```
SolarCrane: Since we are a core action we don't need to create a GUID and a comprehensive name may be better

CoreScriptsRoot/Modules/VR/Recenter.lua
kyle-emmerich: It is, isn't it? Line 77.

SolarCrane: Ah sorry, github for some reason folded the code for 1 line, which, happened to be the one with `spawn(function()` on it

PR #497, opened 2016-08-15
Fix small nil value error

PhoenixShay:
Fix "Panel3D:799 attempt to index field part (a nil value)"


kyle-emmerich:
Hey, thanks for the PR but this should already be fixed in the next release. https://github.com/ROBLOX/Core-Scripts/pull/472

PR #498, opened 2016-08-16
Add Http module in common directory

jmargh:
Adding new Http module to reduce the amount of code needed when making Roblox endpoint calls in CoreScripts. The module wraps all http API calls in pcalls, which is where most of the code redundancy comes from. Each call will decode the json into a lua table, and I have added an optional returnRaw parameter if you need the raw json.

This has been updated and added to the new Common directory. This will not conflict with the Xbox app http module (we may want to reorganize that project to reduce future conflicts). Module has been added to Common as it has use on both the server and the client.

After this is approved, we can start migrating existing calls to use this module.

### DIFF
--- a/CoreScriptsRoot/Modules/Common/Http.lua
+++ b/CoreScriptsRoot/Modules/Common/Http.lua
@@ -1,0 +1,117 @@
+--[[
+                // Http.lua
+
+                // API for making http calls into roblox endpoints
+]]
+local HttpService = game:GetService('HttpService')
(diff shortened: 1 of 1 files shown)


jmargh:
Will leave open for review and feedback, but will need to update directory path when @spotco makes his changes to allow modules to be loaded on the server, as we'll want this on both the client and server.

jmargh:
Request has been updated to reflect new module directories.

SolarCrane:
Looks good to me

PR #499, opened 2016-08-16
CommonModules and example

spotco:
CommonModules


kyle-emmerich:
Might this sort of be the same idea as what's going on here? https://github.com/ROBLOX/Core-Scripts/pull/498

If I remember correctly, Shared was meant to be more like "shared between platforms," but it makes sense to me to be "shared between client and server" as well.

spotco:
thanks worked it out

PR #503, opened 2016-08-18
Fix VR settings menu input, fix dropdown menus in VR, fix alerts in VR

kyle-emmerich:
Waiting to test mobile to make sure nothing has broken. Also going to flag everything. It's a complicated change so I'm making a PR early.

### DIFF
--- a/CoreScriptsRoot/Modules/Settings/Utility.lua
+++ b/CoreScriptsRoot/Modules/Settings/Utility.lua
@@ -9,6 +9,10 @@
 local SELECTED_COLOR = Color3.new(0,162/255,1)
 local NON_SELECTED_COLOR = Color3.new(78/255,84/255,96/255)
 
+local ARROW_COLOR = Color3.new(0.8, 0.8, 0.8)
+local ARROW_COLOR_HOVER = Color3.new(1, 1, 1)
+local ARROW_COLOR_TOUCH = ARROW_COLOR_HOVER
+
 local SELECTED_LEFT_IMAGE = "rbxasset://textures/ui/Settings/Slider/SelectedBarLeft.png"
 local NON_SELECTED_LEFT_IMAGE = "rbxasset://textures/ui/Settings/Slider/BarLeft.png"
 local SELECTED_RIGHT_IMAGE = "rbxasset://textures/ui/Settings/Slider/SelectedBarRight.png"
@@ -211,7 +215,7 @@ local function isTenFootInterface()
 end
 
 local function usesSelectedObject()
-	if UserInputService.TouchEnabled and not UserInputService.GamepadEnabled then return false end
+	if UserInputService.TouchEnabled and not UserInputService.GamepadEnabled or UserInputService.VREnabled then return false end
 
 	return true
 end
@@ -447,6 +451,7 @@ local function MakeButton(name, text, size, clickFunc, pageRef, hubRef)
 end
 
 local function CreateDropDown(dropDownStringTable, startPosition, settingsHub)
+	local Panel3D = require(CoreGui.RobloxGui.Modules.VR.Panel3D)
 	-------------------- CONSTANTS ------------------------
 	local DEFAULT_DROPDOWN_TEXT = "Choose One"
 	local SCROLLING_FRAME_PIXEL_OFFSET = 25
@@ -493,6 +498,21 @@ local function CreateDropDown(dropDownStringTable, startPosition, settingsHub)
 		Parent = CoreGui.RobloxGui
 	};
 
+	local function onVREnabled(prop)
+		if prop ~= "VREnabled" then
+			return
+		end
+		if UserInputService.VREnabled then
+			DropDownFullscreenFrame.Parent = Panel3D.Get("SettingsMenu"):GetGUI()
+			DropDownFullscreenFrame.BackgroundTransparency = 1
+		else
+			DropDownFullscreenFrame.Parent = CoreGui.RobloxGui
+			DropDownFullscreenFrame.BackgroundTransparency = 0.2
+		end
+	end
+	UserInputService.Changed:connect(onVREnabled)
+	onVREnabled("VREnabled")
+
 	local DropDownSelectionFrame = Util.Create'ImageLabel'
 	{
 		Name = "DropDownSelectionFrame",
@@ -535,6 +555,10 @@ local function CreateDropDown(dropDownStringTable, startPosition, settingsHub)
 
 		dropDownButtonEnabled.Value = interactable
 		active = false
+
+		if UserInputService.VREnabled then
+			Panel3D.Get("SettingsMenu"):SetSubpanelDepth(DropDownFullscreenFrame, 0)
+		end
 	end
 	local noOpFunc = function() end
 
@@ -545,6 +569,9 @@ local function CreateDropDown(dropDownStringTable, startPosition, settingsHub)
 		active = true
 
 		DropDownFullscreenFrame.Visible = true
+		if UserInputService.VREnabled then
+			Panel3D.Get("SettingsMenu"):SetSubpanelDepth(DropDownFullscreenFrame, 0.5)
+		end
 		if not this.CurrentIndex then this.CurrentIndex = 1 end
 		if this.CurrentIndex <= 0 then this.CurrentIndex = 1 end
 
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -211,9 +217,7 @@ local function isTenFootInterface()
 end
 
 local function usesSelectedObject()
-	if UserInputService.TouchEnabled and not UserInputService.GamepadEnabled then return false end
-
-	return true
+	return not (UserInputService.TouchEnabled and not UserInputService.GamepadEnabled or UserInputService.VREnabled)
```
SolarCrane: This statement is starting to get a little difficult to parse and you need to think about the order of operations with `or` and `and` statements. Do you think we can add some more parens to make it more obvious?

kyle-emmerich: I have to agree; this file seems to be full of difficult-to-parse logical statements. I'm in favor of splitting this up until multiple lines that spell out what's good and what isn't, even if it's ever so slightly more code.

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -313,6 +317,17 @@ local function isGuiVisible(gui, debug) -- true if any part of the gui is visibl
 	end
 end
 
+local function addHoverState(button, instance, applyNormal, applyHover)
+	local function indirectNormal()	applyNormal(instance) end
```
SolarCrane: The naming on these funcs are a bit confusing for me, I am not sure why it is called `indirectNormal` also I think `applyNormal` might want to be called `onNormalButtonState` or `onButtonNormalState` to disambiguate for the reader (remember 3d guis in VR have surface normals!)

kyle-emmerich: This is true. The intent with "indirectNormal" was that the function provided indirection and can be used for event connections. I can't think of a better name for it...

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1130,25 +1243,46 @@ local function CreateSelector(selectionStringTable, startPosition)
 end
 
 local function ShowAlert(alertMessage, okButtonText, settingsHub, okPressedFunc, hasBackground)
-	if CoreGui.RobloxGui:FindFirstChild("AlertViewFullScreen") then return end
+	local parent = CoreGui.RobloxGui
+	local AlertViewBacking = nil
```
SolarCrane: how come capital 'A' here?

kyle-emmerich: That's how I found it! It looks like all of the "important" layout gui objects are like this.

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -493,6 +517,23 @@ local function CreateDropDown(dropDownStringTable, startPosition, settingsHub)
 		Parent = CoreGui.RobloxGui
 	};
 
+	local function onVREnabled(prop)
+		if prop ~= "VREnabled" then
+			return
+		end
+		if UserInputService.VREnabled then
+			DropDownFullscreenFrame.Parent = Panel3D.Get("SettingsMenu"):GetGUI()
+			DropDownFullscreenFrame.BackgroundTransparency = 1
```
SolarCrane: Might want to save these numbers in a const variable

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -493,6 +517,23 @@ local function CreateDropDown(dropDownStringTable, startPosition, settingsHub)
...
+		else
+			DropDownFullscreenFrame.Parent = CoreGui.RobloxGui
+			DropDownFullscreenFrame.BackgroundTransparency = 0.2
```
SolarCrane: Same as above

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1012,7 +1104,11 @@ local function CreateSelector(selectionStringTable, startPosition)
 					this.Selections[this.CurrentIndex].TextTransparency = 0
 				else
 					if GuiService.SelectedCoreObject ~= nil and isAutoSelectButton[GuiService.SelectedCoreObject] then
-						GuiService.SelectedCoreObject = this.SelectorFrame
+						if UserInputService.VREnabled and fixSettingsMenuVR then
+							this.Selections[this.CurrentIndex].TextTransparency = 0
```
SolarCrane: Do we need to restore the texttransparency here when we disable VR if we ever do in the future?

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1058,6 +1154,25 @@ local function CreateSelector(selectionStringTable, startPosition)
 	end
 
 	--------------------- SETUP -----------------------
+	local function onVREnabled(prop)
+		if prop ~= "VREnabled" then
+			return
+		end
+		if UserInputService.VREnabled then
```
hockeychaos: why not just set Selectable to UserInputService.VREnabled?

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -447,6 +470,7 @@ local function MakeButton(name, text, size, clickFunc, pageRef, hubRef)
 end
 
 local function CreateDropDown(dropDownStringTable, startPosition, settingsHub)
+	local Panel3D = require(CoreGui.RobloxGui.Modules.VR.Panel3D)
```
hockeychaos: this feels weird to require for every drop down, even if we aren't a 3d renderer. Could we just use it if we need it?

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1546,51 +1725,101 @@ local function CreateNewSlider(numOfSteps, startStep, minStep)
...
+			mouseDownFunc(inputObject, i) 
+		end)
+		this.Steps[i].InputEnded:connect(function(inputObject)
+		 mouseUpFunc(inputObject) end)
 	end
 
-	this.SliderFrame.InputBegan:connect(function(inputObject) mouseDownFunc(inputObject, currentStep) end)
-	this.SliderFrame.InputEnded:connect(function(inputObject) mouseUpFunc(inputObject) end)
+	this.SliderFrame.InputBegan:connect(function(inputObject) 
+		if fixSettingsMenuVR then
```
hockeychaos: I might be wrong about this, but I feel like this might break being able to click with a mouse to the specified cell you want. Like for volume, right now you can click which cell you want the volume at.

kyle-emmerich: I think you may be right, although clicking does still work. This just breaks being able to click anywhere in the slider frame to change the value (rather than exactly on a cell.) I'll make this VR-only.

CoreScriptsRoot/Modules/Settings/Utility.lua
kyle-emmerich: I think it gets set somewhere else anyway, this just reproduces the functionality for VR since mouse hover events don't quite work right.

That reminds me, we should see if we can get those working some time and remove these special cases.

CoreScriptsRoot/Modules/Settings/Utility.lua
kyle-emmerich: Did you mean to comment somewhere else?

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1058,6 +1158,25 @@ local function CreateSelector(selectionStringTable, startPosition)
 	end
 
 	--------------------- SETUP -----------------------
+	local function onVREnabled(prop)
+		if prop ~= "VREnabled" then
+			return
+		end
+		if UserInputService.VREnabled then
```
hockeychaos: meant to comment here

CoreScriptsRoot/Modules/Settings/Utility.lua
hockeychaos: that was weird...

kyle-emmerich: are you okay

CoreScriptsRoot/Modules/Settings/Utility.lua
kyle-emmerich: (I did this btw)

PR #507, opened 2016-08-23
New Chat System Bug Fix #2

JacobLocke:
Tweening and message object creation now waits until Gui window has been properly parented to PlayerGui. There's like a single frame during respawn that sets parent to nil.

Also, server does not need to wait for SFFlags to sync.


CoreScriptsRoot/ServerCoreScripts/ClientChat/ChannelsBar.lua
```diff
@@ -281,6 +283,12 @@ function methods:CreateTweeners()
 	
 end
 
+function methods:WaitUntilParentedCorrectly()
```
hockeychaos: why is this necessary? I'm not a huge fan of this method having to be repeated for every file. Are the sizes of the guis wrong when the script is first run? Do we do something similar in the current chat script?

JacobLocke: The way this new chat system has to do persist through death is kind of hacky where it parents everything to nil, then waits for the character to load, then parents everything back to PlayerGui. Because of this, there is at minimum a single frame where there could be messages that try to get put into the message log or channel tabs that try to be created. When they try to do size related stuff when they are not descendants of PlayerGui they will always return a size of 0. This causes messages to get bunched together and overlap. It also breaks the channel bar because it tries to do tweening to get all the tabs ordered and it will throw an error if it's not parented correctly.

hockeychaos: Is there no way to make this function once? I'm worried someday either we will need to tweak it or change something about it and someone will forget to update one of these.

PR #509, opened 2016-08-25
New Chat System Bug Fix / New Features #3

JacobLocke:
Header comments on all files.

Draggable and Resizable config properties now work. You can also set config properties at runtime and have them update such as size of ChannelsBar and TextBar.

Server ChatChannel chat log history. Chat history gets sent to player when they join a channel.


CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatChannel.lua
```diff
@@ -199,11 +207,33 @@ function methods:UnregisterProcessCommandsFunction(funcId)
...
 		speaker:SendSystemMessage(message, self.Name)
 	end
 end
 
+function methods:RemoveExcessMessagesFromLog()
+	while (#self.ChatHistory > self.MaxHistory) do
+		table.remove(self.ChatHistory, 1)
+	end
+end
+
```
BinaryResolved: Instead of calling `table.remove` each time, we could just call it once then call the variable. It can save a quite of CPU usage. So something like this,

> 
function methods:RemoveExcessMessagesFromLog()
local remove = table.remove
while (#self.ChatHistory > self.MaxHistory) do
remove(self.ChatHistory, 1) 
end
>

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatServiceRunner.lua
```diff
@@ -241,6 +251,9 @@ if modules then
 	end
 end
 
+wait()
+wait()
+
```
BinaryResolved: Are two `wait()`'s needed?

JacobLocke: It's to wait 2 frames to give the chat modules, which a bunch of different threads are spawned for, to run and initialize before creating players to put in channels.

I'll put a comment about it in the source file.

CoreScriptsRoot/ServerCoreScripts/ServerChat/ChatServiceRunner.lua
```diff
@@ -241,6 +251,10 @@ if modules then
 	end
 end
 
+--// Wait two frames so the ChatModules have some time to initialize
+wait()
```
spotco: initialize the children in the same thread, and do not wait

spotco: additionally in this case, should specify that module initializations are not expected to wait or yield (or else stuff will break)

PR #511, opened 2016-08-26
New Humanoid Sound script

spotco:
New Humanoid Sound script that's several times more efficient network traffic-wise.
For a networked player in the "worst" case (repeatedly jumping up and down on an upwards slope), reduces network traffic by up to 6x.

TODO: flag loading in cpp code


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+	These sounds are used by the "LocalSound" script.
+]]--
+--humanoidSoundNewLocal.rbxmx--
+
+function CreateNewSound(name, id, looped, pitch)
+	local sound = Instance.new("Sound")
+	sound.SoundId = id
+	sound.Name = name
+	sound.archivable = false
+	sound.Parent = script.Parent.Head
```
hockeychaos: is this script created on the server? I it seems be of Script type, not LocalScript. Is this to allow FE games to hear the sounds of humanoids? Also, should we verify Head exists?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
+<roblox xmlns:xmime="http://www.w3.org/2005/05/xmlmime" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.roblox.com/roblox.xsd" version="4">
+	<External>null</External>
+	<External>nil</External>
+	<Item class="Script" referent="RBXc83eeca86f73463493b2478b693fc06d">
+		<Properties>
+			<bool name="Disabled">false</bool>
+			<Content name="LinkedSource"><null></null></Content>
+			<string name="Name">Sound</string>
+			<string name="ScriptGuid"></string>
+			<ProtectedString name="Source"><![CDATA[--[[
```
hockeychaos: I'd put yourself as author here

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+	FreeFalling = 6;
+	FallingDown = 7;
+	Landing = 8;
+	Splash = 9;
+}
+
+local Humanoid = nil
+local Head = nil
+
+--SFX ID to Sound object
```
hockeychaos: why not just use the strings we used in the server script? or have the server use a similar table

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+	YForLineGivenXAndTwoPts = function(x,pt1x,pt1y,pt2x,pt2y)
+		--(y - y1)/(x - x1) = m
+		local m = (pt1y - pt2y) / (pt1x - pt2x)
+		--float b = pt1.y - m * pt1.x;
+		local b = (pt1y - m * pt1x)
+		return m * x + b
+	end;
+	
+	--Clamps the value of "val" between the "min" and "max"
+	Clamp = function(val,min,max)
```
hockeychaos: we should consider putting this in our math library...

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+		return math.min(max,math.max(min,val))	
+	end;
+	
+	--Gets the horizontal (x,z) velocity magnitude of the given part
```
hockeychaos: this and the one below I think are technically speed actually, since we remove directionality

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+	VerticalVelocityMagnitude = function(Head)
+		return math.abs(Head.Velocity.Y)
+	end;
+	
+	--Checks if Sound.Playing is usable
+	SoundIsV2 = function(sound)
+		return sound.IsLoaded
+	end;
+	
+	--Setting Playing/TimePosition values directly result in less network traffic than Play/Pause/Resume/Stop
```
hockeychaos: Is this is the big thing that saves network traffic? does anything else here? If not, why not just modify current script?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+
+do
+	-- List of all active Looped sounds
+	local activeLoopedSounds = {}
+	
+	-- Last seen Enum.HumanoidStateType
+	local activeState = nil
+	
+	-- Verify and set that the sound's .Looped property is the value of "looped"
+	function verifyAndSetLoopedForSound(sound, looped)
```
hockeychaos: this seems really wordy, why not just setLooped? Also why is this necessary? Internally it looks like we are already doing this check for you

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+		if sound.Looped ~= looped then
+			sound.Looped = looped
+		end
+	end
+	
+	-- Verify and set that "sound" is in "activeLoopedSounds".
+	function setSoundInActiveLooped(sound)
```
hockeychaos: why is this not a part of setLooped?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+		for i=1, #activeLoopedSounds do
+			if activeLoopedSounds[i] == sound then
+				return
+			end
+		end	
+		table.insert(activeLoopedSounds,sound)
+	end
+	
+	-- Stop all active looped sounds except parameter "except". If "except" is not passed, all looped sounds will be stopped.
+	function stopActiveLoopedSoundsExcept(except)
```
hockeychaos: what does the word active add in this context?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+			stateUpdated(Enum.HumanoidStateType.Running)
+		end;
+		
+		[Enum.HumanoidStateType.Running] = function()	
+			local sound = Sounds[SFX.Running]
+			verifyAndSetLoopedForSound(sound,true)
+			stopActiveLoopedSoundsExcept(sound)
+			
+			if Util.HorizontalVelocityMagnitude(Head) > 0.5 then
+				if not sound.IsPlaying then
```
hockeychaos: is this if check really necessary? I don't think it will save much time, and if it does why not just put it in util?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+				splashSound.Volume = Util.Clamp(
+					Util.YForLineGivenXAndTwoPts(
+						Util.VerticalVelocityMagnitude(Head), 
+						100, 0.28, 
+						350, 1),
+					0,1)
+				Util.Play(splashSound)
+			end
+			
+			do
```
hockeychaos: is the do here just for scoping or logical separation?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+			verifyAndSetLoopedForSound(sound,false)
+			Util.Play(sound)
+		end;
+		
+		[Enum.HumanoidStateType.Freefall] = function()
+			if activeState == Enum.HumanoidStateType.Freefall then
+				return
+			end
+			local sound = Sounds[SFX.FreeFalling]
+			if sound.Volume ~= 0 then
```
hockeychaos: why might as well remove this if

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,381 @@
...
+	-- Handle state event fired or OnChange fired
+	function stateUpdated(state)
+		if stateUpdateHandler[state] ~= nil then
+			stateUpdateHandler[state]()
+		end
+		activeState = state
+	end
+	
+	-- Runs on heartbeat
+	function onHeartbeat(step)
```
hockeychaos: can we not do this with events? it seems like it would be more performant to not need a loop here. If a loop is necessary why not spawn or delay a function in the freefall state that does this?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: So it's 4 big savings:
1. Stopping only the active looped sounds (stopActiveLoopedSoundsExcept). Old script would stop all looped sounds (always stop swimming,running and climbing), even if they weren't playing.

2. Use of .Playing and .TimePosition if they're available (instead of Play/Pause/Resume/Stop). The Play methods are more network traffic because they've got additional replication things they need to do (update the internal "lastSetLuaTime", fire Played/Paused/Resumed/Stopped events)

3. Changes to the behaviour of the "Running" sound and "Falling" sound to not trigger quite as often. These are subtle behaviour changes, but what the player hears ingame should be indistinguishable.

4. Avoids any normally un-noticeable changes to Volume/Pitch (there were a couple of not-very-functional speed-to-pitch linear relationships in the old code, they're just set-once pitches now). This is only network traffic in non-filtering enabled games.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: youre right this will be changed

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: this is a little different:
so this localscript maintains a list of "looped sounds" that are actively playing.
this "setSoundInActiveLooped" adds the parameter sound to the "activeLoopedSounds" array.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: this modifies the "activeLoopedSounds" array

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: yeah this one can go into Util

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: logical separation

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: yes

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: ok

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: ill use object.Name

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: where is this

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: name will b changed

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: old code doesnt verify that but it can be done

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
hockeychaos: I just mean the standard lua math lib

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
hockeychaos: yeah it just seems like if we have a function called setLooped we should also call setSoundInActiveLooped if looped == true. Or just remove this function entirely and have it as part of setLooped.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
hockeychaos: the word active is a bit vague to me. Does that mean it is playing? Does it mean it can play? Does it mean it is the main sound that is playing?

spotco: ill rename active -> playing.
It's "an" active looped sound that is playing. For any state they may be one, multiple or none. Depends on the state handler implementation.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: Heartbeat's here for some new behaviour:
1. Only trigger the freefalling sound if the last active state was "FreeFall", and vertical speed is above 75 (arbitrary number that made it NOT trigger for normal jumps). 

2. If last active state was running, and horizontal speed is less than 0.5, stop the running sound playback.

I personally think it's a lot cleaner to do this with a stateless update method (or else we'd need to manage/preempt the spawn or delay functions). The only "state" that this code has right now is "lastActiveState" and "activeLoopedSounds" (soon to be renamed playingLoopedSounds).

And they don't seem to be a big problem performance-wise. Heartbeat stuff generally shouldn't be, is that correct? (It's one heartbeat listener per client too)

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: Thinking about it:
This is to keep the SFX.FreeFalling syntax (as opposed to SFX["FreeFalling"]).
The first would be better in a typed language because of compiler checking, I'm guessing the errors in lua would be about the same.
Anyway, I'm now for keeping it as is (SFX.FallingDown) because easier to read?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
spotco: re: I'll do another PR sometime with some more math_whatever functions.
I see lua_pushnumber and immediately go nope.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
hockeychaos: also remember that this has to run on a mobile device in VR, so everything counts.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -0,0 +1,382 @@
...
+		return math.min(max,math.max(min,val))	
+	end;
+	
+	--Gets the horizontal (x,z) velocity magnitude of the given part
+	HorizontalSpeed = function(Head)
+		local hVel = Head.Velocity + Vector3.new(0,-Head.Velocity.Y,0)
+		return hVel.magnitude	
+	end;
+	
+	TickSpeedSecondsAndLengthSecondsToNormalizedIncrement = function(tickSpeedSeconds, lengthSeconds)
```
hockeychaos: Is this function necessary? The name is kind of confusing and it is longer than the actual function

spotco: u got a problem with longer names m8?

(Anyway, this math is necessary to keep the update timestep variable, and the unit of the number in the application code in seconds)

hockeychaos: oh I just meant lets just put the code its doing directly into the function, since its just a divide and its the only place we use it.

PR #512, opened 2016-08-26
New Chat System Bug Fix / New Features #4

JacobLocke:
Chat window can no longer be selected by GUI navigation via gamepad.

ChatBar TextBox doesn't break with tweening anymore.


spotco:
Update on this:
Can we leave selectable on for all the relevant elements? (Was somewhat useable on gamepad with it on)

PR #518, opened 2016-08-29
CLI-11825 Improve Abuse Reports from Games

garnold24:
Removed all game abuse reports but inappropriate content
defaults to reporting a player unless you're alone in a server
fixed some logic with the submit button not appearing properly
and made it so game details get sent along with the report message.

### DIFF
--- a/CoreScriptsRoot/Modules/Settings/Pages/ReportAbuseMenu.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/ReportAbuseMenu.lua
@@ -180,7 +180,7 @@ local function Initialize()
 				makeSubmitButtonActive()
 			else
 				this.TypeOfAbuseMode:UpdateDropDownList(ABUSE_TYPES_PLAYER)
-				this.TypeOfAbuseMode:SetInteractable(true)
+				this.TypeOfAbuseMode:SetInteractable(#ABUSE_TYPES_PLAYER > 1)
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/Settings/Pages/ReportAbuseMenu.lua
```diff
@@ -172,11 +171,16 @@ local function Initialize()
...
+				
+				this.TypeOfAbuseMode:SetInteractable(#ABUSE_TYPES_GAME > 1)
+
 				this.WhichPlayerMode:SetInteractable(false)
 				this.WhichPlayerLabel.ZIndex = 1
 				this.GameOrPlayerMode.SelectorFrame.NextSelectionDown = this.TypeOfAbuseMode.DropDownFrame
+				makeSubmitButtonActive()
 			else
 				this.TypeOfAbuseMode:UpdateDropDownList(ABUSE_TYPES_PLAYER)
+				this.TypeOfAbuseMode:SetInteractable(true)
```
hockeychaos: should we do the same thing as for ABUSE_TYPES_GAME?

garnold24: Although it's likely that there won't be just one reason to report another player, we should for consistency.

PR #520, opened 2016-08-29
VR-229 Put the report abuse success alert on a subpanel

kyle-emmerich:


CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -749,6 +750,15 @@ local function CreateDropDown(dropDownStringTable, startPosition, settingsHub)
 		this.Selections = {}
 		this.SelectionInfo = {}
 
+		local vrEnabled = UserInputService.VREnabled
```
SolarCrane: Does this stuff get updated in the `onVREnabled` function or no?

kyle-emmerich: Hm, it's in UpdateDropDownList(). I think this gets called again at a useful time, but I'll double-check that.

PR #521, opened 2016-08-29
VR-177 Manage cursor dot behavior (settings panel doesn't come up wit…

kyle-emmerich:
…h radial menu now and cursor stays hidden until radial menu exit animation finishes)

### DIFF
--- a/CoreScriptsRoot/CoreScripts/GamepadMenu.lua
+++ b/CoreScriptsRoot/CoreScripts/GamepadMenu.lua
@@ -533,14 +533,16 @@ local function setupGamepadControls()
 		end
 	end
 
-	function setOverrideMouseIconBehavior()
-		pcall(function()
+	function setOverrideMouseIconBehavior(override)
+		if override then
 			if InputService:GetLastInputType() == Enum.UserInputType.Gamepad1 then
 				InputService.OverrideMouseIconBehavior = Enum.OverrideMouseIconBehavior.ForceHide
 			else
 				InputService.OverrideMouseIconBehavior = Enum.OverrideMouseIconBehavior.ForceShow
 			end
-		end)
+		else
+			InputService.OverrideMouseIconBehavior = Enum.OverrideMouseIconBehavior.None
+		end
 	end
 
 	function toggleCoreGuiRadial(goingToSettings)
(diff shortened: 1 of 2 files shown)


CoreScriptsRoot/CoreScripts/GamepadMenu.lua
```diff
@@ -581,6 +580,9 @@ local function setupGamepadControls()
 			gamepadSettingsFrame:TweenSizeAndPosition(UDim2.new(0,102,0,102), UDim2.new(0.5,-51,0.5,-51),
 														Enum.EasingDirection.Out, Enum.EasingStyle.Sine, 0.1, true, 
 				function()
+					if not InputService.VREnabled then
+						pcall(function() InputService.OverrideMouseIconBehavior = Enum.OverrideMouseIconBehavior.None end)
```
SolarCrane: Seems like most updates to `OverrideMouseIconBehavior` runs through the `setOverrideMouseIconBehavior()` function.
In order to prevent conflicts between the setting of this property by that function we should run this logic there instead

kyle-emmerich: This is what the old code was doing, but I moved it to make sure the cursor doesn't show for a single frame in VR. The setOverrideMouseIconBehavior() function only shows/hides it, but doesn't reset to none.

SolarCrane: What I mean is we should improve the `setOverrideMouseIconBehavior` function and then invoke that function instead of just setting the mouse behavior here with no regard to the function.

kyle-emmerich: That'd work

CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -977,17 +977,18 @@ local function CreateSettingsHub()
 		this.VRShield.Visible = true
 		this:HideShield()
 
-		vrMenuOpened = GuiService.MenuOpened:connect(function()
-			local topbarPanel = Panel3D.Get("Topbar3D")
-			panel.localCF = topbarPanel.localCF * CFrame.Angles(math.rad(-5), 0, 0) * CFrame.new(0, 4, 0) * CFrame.Angles(math.rad(-15), 0, 0)
-			panel:SetVisible(true)
+		vrMenuOpened = this.SettingsShowSignal:connect(function(visible)
+			if visible and this.Visible then
```
hockeychaos: why do we check this.Visible there? is this if statement equivalent to asking if we are trying to open the settings hub when it is already open?

kyle-emmerich: Yes, that's the intent.

PR #522, opened 2016-08-31
Fixing bug when game settings doesn't show that

BinaryResolved:
...fullscreen was toggled with F11.

### DIFF
--- a/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
@@ -103,11 +103,7 @@ local function Initialize()
 
 		settingsDisabledInVR[this.FullscreenEnabler] = true
 
-		local Debounce = false
-		
 		this.FullscreenEnabler.IndexChanged:connect(function(newIndex)
-			if Debounce then return end
-			Debounce = true
 			if newIndex == 1 then
 				if not GameSettings:InFullScreen() then
 					GuiService:ToggleFullscreen()
@@ -119,12 +115,9 @@ local function Initialize()
 					this.FullscreenEnabler:SetSelectionIndex(2)
 				end
 			end
-			Debounce = false
 		end)
 
 		GameSettings.FullscreenChanged:connect(function(isFullScreen)
-			if Debounce then return end
-			Debounce = true
 			if isFullScreen then
 				if this.FullscreenEnabler:GetSelectedIndex() ~= 1 then
 					this.FullscreenEnabler:SetSelectionIndex(1)
@@ -134,7 +127,6 @@ local function Initialize()
 					this.FullscreenEnabler:SetSelectionIndex(2)
 				end	
 			end
-			Debounce = false
 		end)
 		
 		------------------ Gfx Enabler Selection GUI Setup ------------------
--- a/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
@@ -1,4 +1,3 @@
-
 --[[
 		Filename: GameSettings.lua
 		Written by: jeditkacheff
--- a/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
@@ -103,6 +103,8 @@ local function Initialize()
 
 		settingsDisabledInVR[this.FullscreenEnabler] = true
 
+		local Debounce = false
+		
 		this.FullscreenEnabler.IndexChanged:connect(function(newIndex)
 			if Debounce then return end
 			Debounce = true
--- a/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
@@ -1,3 +1,4 @@
+
 --[[
 		Filename: GameSettings.lua
 		Written by: jeditkacheff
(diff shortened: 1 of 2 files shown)


CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
```diff
@@ -103,10 +103,38 @@ local function Initialize()
...
-
+		local Debounce = false
+		
 		this.FullscreenEnabler.IndexChanged:connect(function(newIndex)
-			GuiService:ToggleFullscreen()
+			if Debounce then return end
+			Debounce = true
+			if newIndex == 1 then
+				if not GameSettings:InFullScreen() then
+					GuiService:ToggleFullscreen()
```
hockeychaos: seems this toggle can pulled out of the if/else and just done at the end

CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
```diff
@@ -103,10 +103,38 @@ local function Initialize()
 
 		settingsDisabledInVR[this.FullscreenEnabler] = true
 
-		local fullScreenSelectionFrame = this.FullscreenEnabler.SliderFrame and this.FullscreenEnabler.SliderFrame or this.FullscreenEnabler.SelectorFrame
-
+		local Debounce = false
```
hockeychaos: I don't think this debounce is necessary, did you have a situation this fixed?

CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
```diff
@@ -103,10 +103,38 @@ local function Initialize()
...
+			elseif newIndex == 2 then
+				if GameSettings:InFullScreen() then
+					GuiService:ToggleFullscreen()
+					this.FullscreenEnabler:SetSelectionIndex(2)
+				end
+			end
+			Debounce = false
+		end)
+
+		GameSettings.FullscreenChanged:connect(function(isFullScreen)
```
hockeychaos: Were you able to test these changes? I don't know if it works exactly like this, but you may be able to leave this.FullscreenEnabler.IndexChanged as it was before, and then just update the gui with this function (I believe ToggleFullscreen will fire this function)

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1040,8 +1040,29 @@ local function CreateSelector(selectionStringTable, startPosition)
 			local isSelected = (i == index)
 
 			if not selectionLabel:IsDescendantOf(game) then
-				this.CurrentIndex = i
-				indexChangedEvent:Fire(index)
+				for i,selectionLabel in pairs(this.Selections) do
```
hockeychaos: I'd prefer to keep a tweening update outside the scope of a bugfix, unless this fixes something I'm not aware of.

CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
BinaryResolved: If I leave the `IndexChanged` function as it was before then it would rapidly toggle the fullscreen on and off. Due to the `GuiService:ToggleFullscreen()` firing the `FullscreenChanged` which then invokes the `IndexChanged` again.

CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
BinaryResolved: The debounce was from another angle I was trying to fix the bug from. I guess I never tested this way without the debounce, seems to work fine without it. I'll remove it.

CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
BinaryResolved: If I take it out of the if/else then the same thing would happen as below. If I left the whole function the same as it was.

CoreScriptsRoot/Modules/Settings/Utility.lua
BinaryResolved: Yes, when `:SetSelectionIndex()` is fired from GameSettings the selectionLabel aren't considered a descendant of game. And throws the error 

> Can only tween objects in the workspace

And that part of the code is only ran when Fullscreen is toggled with F11 not the slider.

hockeychaos: hmm, it seems like this code already handles this case?

`if not selectionLabel:IsDescendantOf(game) then
				this.CurrentIndex = i
				indexChangedEvent:Fire(index)
				return
			end`

If the issue is the gui does not update properly now because we exit out of doing all the ui updates. I think the cleanest solution would be to:

1) get rid of this if check and return
2) do all the normal tweening logic as it is now, but when we get to the tween calls, check if we are a descendantOf(game), if so do actual tweening calls, otherwise just set the positions/properties directly.

BinaryResolved: That was indeed the problem, fixed.

CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1059,13 +1052,21 @@ local function CreateSelector(selectionStringTable, startPosition)
 				selectionLabel.Position = tweenPos
 				selectionLabel.Visible = true
 				PropertyTweener(selectionLabel, "TextTransparency", 1, 0, TweenTime * 1.1, EaseOutQuad)
-				selectionLabel:TweenPosition(UDim2.new(0,leftButton.Size.X.Offset,0,0), Enum.EasingDirection.In, Enum.EasingStyle.Quad, TweenTime, true)
+				if selectionLabel:IsDescendantOf(game) then
+					selectionLabel:TweenPosition(UDim2.new(0,leftButton.Size.X.Offset,0,0), Enum.EasingDirection.In, Enum.EasingStyle.Quad, TweenTime, true)
```
hockeychaos: one last nitpicky thing, can we define the UDim2 before the if/else? Don't want someone to accidentally only update one of these in the future

PR #525, opened 2016-09-06
Primitive UI for Quick Profiler

DougBanksPersonal:
Data models to keep a running history of key stats.

Buttons to show stats in a small button

An "Enlarged view" that toggles on/off with button clicks.

Stubbed out "graph".


CoreScriptsRoot/Modules/Panel3D.lua
```diff
@@ -0,0 +1,491 @@
+--Panel3D: 3D GUI panels for VR
```
hockeychaos: I have a feeling you didn't want to add this? I think it was removed at some point

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
```diff
@@ -0,0 +1,6 @@
+--[[ Services ]]--
+local CoreGuiService = game:GetService('CoreGui')
```
hockeychaos: is this going to be updated at some point? would rather not add if not necessary

CoreScriptsRoot/StarterScript.lua
```diff
@@ -20,12 +20,15 @@ local luaControlsSuccess, luaControlsFlagValue = pcall(function() return setting
...
 
--- MainBotChatScript (the Lua part of Dialogs)
+-- MainBotChatScript (the Lua part of Dialogs)
 scriptContext:AddCoreScriptLocal("CoreScripts/MainBotChatScript2", RobloxGui)
 
 -- In-game notifications script
 scriptContext:AddCoreScriptLocal("CoreScripts/NotificationScript2", RobloxGui)
 
+-- Performance Stats Management
+scriptContext:AddCoreScriptLocal("CoreScripts/PerformanceStatsManagerScript", RobloxGui)
```
hockeychaos: this is a rather large change, so I'd like to flag it here, instead of inside the code just in case something happens from the time the flag is read until we do something about it

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,219 @@
...
+end
+
+--[[ Functions ]]--
+
+-- Set localPlayer to value of Players.localPlayer
+-- Update our screenGui parent accordingly.
+function UpdateLocalPlayer()
+  localPlayer = PlayersService.LocalPlayer
+  if localPlayer then
+    screenGui.Parent = localPlayer:WaitForChild("PlayerGui")
```
hockeychaos: do we need to parent these to the player gui? In general anything in core scripts should be parented to RobloxGui inside of CoreGuiService

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,219 @@
...
+end
+ 
+ 
+-- Set up our GUI.
+ConfigureMasterFrame()
+ConfigureStatButtonsInMasterFrame()
+ConfigureStatViewerInMasterFrame()
+
+-- Watch for changes in local player.
+PlayersService.PlayerAdded:connect(UpdateLocalPlayer)
```
hockeychaos: local player never changes, should not be necessary for this

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,219 @@
...
+  else
+    screenGui.Parent = nil
+  end
+end
+
+function ConfigureMasterFrame()
+  -- Set up the main frame that contains the whole PS GUI.  
+	masterFrame.Position = UDim2.new(0, 0, 0, 0)
+	masterFrame.Size = UDim2.new(1, 0, 1, 0)
+  masterFrame.Draggable = false
```
hockeychaos: this is the default value btw

CoreScriptsRoot/Modules/Stats/StatsButton.lua
```diff
@@ -0,0 +1,106 @@
...
+  print ("SetIsSelected ", self._type)
+  self._isSelected = isSelected
+  self:_updateColor();
+end
+
+function StatsButtonClass:_updateColor()
+  StatsUtils.StyleButtonSelected(self._button, self._isSelected)  
+end
+
+function StatsButtonClass:SetGUIParent(parent)
```
hockeychaos: I'd prefer if you just called this SetParent, as it doesn't need to be a gui

CoreScriptsRoot/Modules/Stats/StatsAggregator.lua
```diff
@@ -0,0 +1,130 @@
...
+  self._listeners[listenerId] = nil
+end
+
+function StatsAggregatorClass:_notifyAllListeners()
+  for key, value in pairs(self._listeners) do
+    value()
+  end
+end
+
+function StatsAggregatorClass:StartListening()
```
hockeychaos: there is no way to kill this thread? Seems like we might want a way to turn this on/off depending on whether we are showing the stats, or at least just a method to turn off stat tracking

CoreScriptsRoot/Modules/Stats/StatsButton.lua
```diff
@@ -0,0 +1,106 @@
...
+  self._isSelected = false
+  
+  self:_updateColor();
+  
+  return self
+end
+
+function StatsButtonClass:SetToggleCallbackFunction(callbackFunction) 
+    self._button.MouseButton1Click:connect(function() 
+          print("Button clicked!!!")
```
hockeychaos: we should remove this before turning the flag on

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
DougBanksPersonal: Yes, it's a stub.  Working on filling it out now.

CoreScriptsRoot/StarterScript.lua
DougBanksPersonal: Done

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
DougBanksPersonal: I don't understand the comment "parent these to the player gui":
1) what are "these": functions, files, ???
2) what does it mean to 'parent' something?

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
DougBanksPersonal: I thought you don't necessarily have a local player right away when you join.  Don't you need to wait for it?

CoreScriptsRoot/Modules/Stats/StatsAggregator.lua
DougBanksPersonal: I can add a method but I am not sure when it'd be called.  

When I first go from "not showing stats" to "showing stats" it'd be nice to show a full graph of the past N measurements -> I have to aggregate even when stats aren't showing.  If it becomes a performance hit we could punt on that and just start collecting when stats are on.

But I'm only collecting once per second (or half second, something like that) so I can't imagine this is that expensive.

hockeychaos: Don't think it'll be a big hit, but it does mean we will have to be careful about an ever growing list of measurements, might want to eventually cull some old ones.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
hockeychaos: specifically talking about screenGui.Parent. So we create this screenGui and then parent it to localPlayer.PlayerGui. PlayerGui is accessible by developer scripts, its also where they would put their own 2d UI for their game. We have a separate container for roblox created UI. Developers can't access this, and it renders on top of all developer UI. Looks like above you grab a reference to this container via CoreGuiService:WaitForChild("RobloxGui"). I'd parent this screenGui to that container instead.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
hockeychaos: it might be possible, but in that case I'd still get rid of these two connections to UpdateLocalPlayer. Instead, I'd do this below, where we explicitly call UpdateLocalPlayer:

` localPlayer = PlayersService.LocalPlayer
  while not localPlayer do
     PlayersService.Changed:wait()
     localPlayer = PlayersService.LocalPlayer
  end`

This way we don't have to listen to extraneous events that won't ever change the thing you are interested in

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,184 @@
...
+
+--[[ Services ]]--
+local PlayersService = game:GetService("Players")
+local Settings = UserSettings()
+local GameSettings = Settings.GameSettings
+local CoreGuiService = game:GetService('CoreGui')
+
+--[[ Modules ]]--
+local RobloxGuiFolder = CoreGuiService:WaitForChild("RobloxGui")
+local ModulesFoler = RobloxGuiFolder:WaitForChild("Modules")
```
hockeychaos: minor, but spelling

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,184 @@
...
+if not showPerformanceStatsInGui then 
+	return
+end
+
+-- Set up our GUI.
+ConfigureMasterFrame()
+ConfigureStatButtonsInMasterFrame()
+ConfigureStatViewerInMasterFrame()
+
+PlayersService.PlayerAdded:connect(UpdatePerformanceStatsVisibility)
```
hockeychaos: shouldn't need the player added/removed events for this, just wait until local player is not nil then update once.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
DougBanksPersonal: Fixed.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
DougBanksPersonal: I want the Stats stuff to show when we are "playing" and not show when we're just in Studio.
This seems necessary/sufficient for that behavior.

hockeychaos: I don't think it is. Since we restart the dm when you transition from edit->play and vice versa, PlayerAdded/Removed should never get called in edit mode. But in an online game/testing play mode in studio those functions will be called often (any time any other play joins/leaves the game), but are never necessary after LocalPlayer is initialized once, since it never changes during a play session.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,191 @@
...
+--[[ Services ]]--
+local PlayersService = game:GetService("Players")
+local Settings = UserSettings()
+local GameSettings = Settings.GameSettings
+local CoreGuiService = game:GetService('CoreGui')
+
+--[[ Modules ]]--
+local RobloxGuiFolder = CoreGuiService:WaitForChild("RobloxGui")
+local ModulesFolder = RobloxGuiFolder:WaitForChild("Modules")
+local StatsFolder = ModulesFolder:WaitForChild("Stats")
```
SolarCrane: If they are packaged with the client we do not have to WaitForChild here, in fact it sometimes hides bugs such as missing files.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,191 @@
...
+--[[ Fast Flags ]]--
+local getShowPerformanceStatsInGuiSuccess, showPerformanceStatsInGuiValue = 
+	pcall(function() return settings():GetFFlag("ShowPerformanceStatsInGui") end)
+local showPerformanceStatsInGui = getShowPerformanceStatsInGuiSuccess and showPerformanceStatsInGuiValue
+
+
+--[[ Script Variables ]]--
+local masterFrame = Instance.new("Frame")
+masterFrame.Name = "PS_MasterFrame"
+local localPlayer = PlayersService.LocalPlayer
```
SolarCrane: Sometimes localplayer hasn't been filed out yet

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,191 @@
...
+  statsButtonsByType[displayType] = button
+end
+
+--[[ Functions ]]--
+function ConfigureMasterFrame()
+  -- Set up the main frame that contains the whole PS GUI.  
+  -- Avoid the top button bar.
+	masterFrame.Position = UDim2.new(0, 0, 0, 0)
+	masterFrame.Size = UDim2.new(1, 0, 1, 0)
+  masterFrame.Selectable = false
```
SolarCrane: Why do we indent back here?

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
```diff
@@ -0,0 +1,191 @@
...
+  -- Set up the widget that shows currently selected button.
+  statsViewer:SetParent(masterFrame)
+  
+  local size = UDim2.new(0.5, 0, 0.5, 0)
+  local position = UDim2.new(0.5, 0, 0.25, 0)
+  statsViewer:SetSizeAndPosition(size, position)
+end
+
+function UpdatePerformanceStatsVisibility() 
+  local localPlayer = PlayersService.LocalPlayer
```
SolarCrane: We create a localPlayer variable at the top of this script. Do we need that variable if we are just fetching it here

CoreScriptsRoot/Modules/Stats/AllStatsAggregators.lua
```diff
@@ -0,0 +1,58 @@
...
+		Filename: AllStatsAggregators.lua
+		Written by: dbanks
+		Description: Indexed array of stats aggregators, one for each stat.
+--]]
+
+--[[ Services ]]--
+local CoreGuiService = game:GetService('CoreGui')
+
+--[[ Modules ]]--
+local folder = CoreGuiService:WaitForChild("RobloxGui")
```
SolarCrane: Probably don't need to WaitForChild here

CoreScriptsRoot/Modules/Stats/StatsAggregator.lua
```diff
@@ -0,0 +1,134 @@
...
+  self._listeners[id] = callbackFunction
+  return id
+end
+
+function StatsAggregatorClass:RemoveListener(listenerId)
+  self._listeners[listenerId] = nil
+end
+
+function StatsAggregatorClass:_notifyAllListeners()
+  for key, value in pairs(self._listeners) do
```
SolarCrane: Can we get better names than key and value

CoreScriptsRoot/Modules/Stats/AllStatsAggregators.lua
```diff
@@ -0,0 +1,58 @@
+
+--[[
+		Filename: AllStatsAggregators.lua
```
SolarCrane: Seems like the typical naming pattern would yield: StatsAggregatorManager

CoreScriptsRoot/Modules/Stats/StatsViewer.lua
```diff
@@ -0,0 +1,123 @@
...
+      text and graphics.
+--]]
+
+--[[ Services ]]--
+local CoreGuiService = game:GetService('CoreGui')
+
+--[[ Modules ]]--
+local folder = CoreGuiService:WaitForChild("RobloxGui")
+folder = folder:WaitForChild("Modules")
+folder = folder:WaitForChild("Stats")
```
SolarCrane: What is the deal here? Can we do this all in one line?

DougBanksPersonal: Removed.

CoreScriptsRoot/Modules/Stats/AllStatsAggregators.lua
DougBanksPersonal: Renamed.

CoreScriptsRoot/Modules/Stats/StatsAggregator.lua
DougBanksPersonal: Yes!

CoreScriptsRoot/Modules/Stats/AllStatsAggregators.lua
DougBanksPersonal: Removed.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
DougBanksPersonal: Done.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
DougBanksPersonal: Maybe, think I fixed it.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
DougBanksPersonal: Removed.

CoreScriptsRoot/CoreScripts/PerformanceStatsManagerScript.lua
DougBanksPersonal: removed.

PR #527, opened 2016-09-08
Add bar chart for performance stats.

DougBanksPersonal:


CoreScriptsRoot/Modules/Stats/BarGraph.lua
```diff
@@ -0,0 +1,62 @@
...
+--]]
+
+--[[ Services ]]--
+local CoreGuiService = game:GetService('CoreGui')
+
+--[[ Globals ]]--
+local BarColor = Color3.new(0.1, 0.7, 0.1)
+
+--[[ Modules ]]--
+local folder = CoreGuiService:WaitForChild("RobloxGui")
```
SolarCrane: This line could be `local folder = CoreGuiService.RobloxGui.Modules.Stats` and be fine

hockeychaos: Might be late to the party on this one as I think we call this folder in a lot of other places, but a name like StatsFolder may be more descriptive

CoreScriptsRoot/Modules/Stats/BarGraph.lua
```diff
@@ -0,0 +1,62 @@
...
+  return self
+end
+
+function BarGraphClass:PlaceInParent(parent, size, position) 
+  self._frame.Position = position
+  self._frame.Size = size
+  self._frame.Parent = parent
+end
+
+function BarGraphClass:Render(values, axisMax)
```
hockeychaos: Could we call this update or something similar? Render feels a little misleading, like I would need to call this on every render step for it to function correctly.

CoreScriptsRoot/Modules/Stats/BarGraph.lua
```diff
@@ -0,0 +1,62 @@
...
+function BarGraphClass:Render(values, axisMax) 
+  self._frame:ClearAllChildren()
+  
+  local numValues = table.getn(values)
+  for i, value in ipairs(values) do
+    self:_addBar(i, value, numValues, axisMax)
+  end
+end
+
+function BarGraphClass:_addBar(i, value, numValues, axisMax)
```
hockeychaos: is there a reason this function starts with an underscore but the rest don't?

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
```diff
@@ -9,42 +9,164 @@
...
   self._frame.Position = position
   self._frame.Size = size
   self._frame.Parent = parent
 end
 
 function StatsAnnotatedGraphClass:SetValues(values)
+  local axisMax = self:_calculateAxisMax(values)
+  self._graph:Render(values, axisMax)
+  
   -- FIXME(dbanks)
```
hockeychaos: still need this here?

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
```diff
@@ -9,42 +9,164 @@
...
-  -- Fill this in.
+  -- Format based on stat type.
+  self._topLabel.Text = string.format("%.4f", axisMax)
+  self._midLabel.Text = string.format("%.4f", axisMax/2)
+  self._bottomLabel.Text = string.format("%.4f", 0,.0)
+end
+
+function StatsAnnotatedGraphClass:_calculateAxisMax(values)
```
hockeychaos: similar to BarGraph, reason for underscore?

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
```diff
@@ -9,42 +9,164 @@
...
+  if (self._aggregator ~= nil) then
+    self._listenerId = aggregator:AddListener(function()
+        self:_updateValue()
+    end)
+  end
+  
+  self:_updateValue()
+end
+
+function StatsAnnotatedGraphClass:_recursiveGetAxisMax(axisMax, max)
```
hockeychaos: same underscore ?

CoreScriptsRoot/Modules/Stats/BarGraph.lua
DougBanksPersonal: SolarCrane: done.  History from the thought-I-needed-to-wait.
hockeychaos: not an issue anymore, variable is gone.

CoreScriptsRoot/Modules/Stats/BarGraph.lua
DougBanksPersonal: Done

CoreScriptsRoot/Modules/Stats/BarGraph.lua
DougBanksPersonal: Underscore + no caps = "private" or "helper function".
No underscore + caps = "Public".

Is there another convention I should use?

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
DougBanksPersonal: Yes.  Right now I just read raw numbers out of stats and render them to screen.  But things are more readable if I can put in units/formatting based on stat type (so Network stats look like 1.2 Kb/s instead of 1223.002)

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
DougBanksPersonal: See above.

DougBanksPersonal:
Yes.

CoreScriptsRoot/Modules/Stats/BarGraph.lua
hockeychaos: hmm, they aren't actually private though right? I think you could just define local functions inside the new function if you want actual private functions

DougBanksPersonal: No, not really private.  Just notation/readability.

"just define local functions inside the new() function"
Not sure what you mean by that.  I don't so much *need* them to be private, I'd rather keep the code readable than have weird functions-inside-functions.

hockeychaos: ah, yeah I just tried this and it gets messy. I think this is good enough for now.

PR #533, opened 2016-09-09
CLI-13103 Opening the Escape menu doesn't stop the camera from panning

garnold24:
When the menu opens, the camera should stop panning, and on mobile, touch events will be ended, fixing an issue that caused panning to break.


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -974,6 +979,28 @@ local function CreateCamera()
...
+			OnMousePanButtonReleased() -- this function doesn't seem to actually need parameters
+			
+			if UserInputService.TouchEnabled then
+				--[[menu opening was causing serious touch issues
+				this should disable all active touch events if 
+				they're active when menu opens.]]
+				for inputObject, value in pairs(fingerTouches) do
+					fingerTouches[inputObject] = nil
+				end
+				panBeginLook = nil
```
hockeychaos: Should this be a standard reset function? Seems like we should just clear all input?

Issue #536, opened 2016-09-13
attempt to index upvalue 'outputMessageSync'


einsteinK:
```python
21:34:52.31840 (1618.9503372) Error:   CoreGui.RobloxGui.Modules.DeveloperConsoleModule:781: attempt to index upvalue 'outputMessageSync' (a nil value)
21:34:52.31840 (1618.9503372)          Stack Begin
21:34:52.31840 (1618.9503372)          Script 'CoreGui.RobloxGui.Modules.DeveloperConsoleModule', Line 781 - field OpenCallback
21:34:52.31840 (1618.9513379)          Script 'CoreGui.RobloxGui.Modules.DeveloperConsoleModule', Line 2241 - method SetOpen
21:34:52.31840 (1618.9513379)          Script 'CoreGui.RobloxGui.Modules.DeveloperConsoleModule', Line 845 - field new
21:34:52.31840 (1618.9513379)          Script 'CoreGui.RobloxGui.Modules.DeveloperConsoleModule', Line 2998 - upvalue getDeveloperConsole
21:34:52.31840 (1618.9523386)          Script 'CoreGui.RobloxGui.Modules.DeveloperConsoleModule', Line 3009 - method GetVisibility
21:34:52.31840 (1618.9523386)          Script 'CoreGui.RobloxGui.Modules.Settings.SettingsHub', Line 516
21:34:52.31840 (1618.9523386)          Stack End
```
Happens when pressing F9 quickly several times.
Opening the developer console will work, but the local log won't be created yet.
Here's a screenshot: http://prntscr.com/cho2nf/direct

Seems like a racing condition.
To replicate in studio:

* Insert the DeveloperConsoleModule.lua as a script
* At line 2748 add a wait(0.1) *(custom lag, as the HTTP above it errors when not online)*
* At the bottom of the script, before the return statement, add:
```lua
spawn(function()
	DevConsoleModuleTable:GetVisibility()
end)
spawn(function()
	DevConsoleModuleTable:SetVisibility(true)
end)
```
* In the commandline, do:
```lua
loadstring(workspace.TheScript.Source)()
```
* Observe...

I might fix this in the next 10m or tomorrow if I don't have enough time.

GitBlox:
This issue was moved to GitBlox/Core-Scripts#25

ArceusInator:
Haw Can I Eat UR PooPoo Doodoo Crust??????? Hur.


[...]

PR #537, opened 2016-09-13
Fixing issue #536

einsteinK:
Fixing issue #536 was easier than I expected.
This'll make sure only one console will be created, using a sort of debounce.

### DIFF
--- a/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
+++ b/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
@@ -2992,6 +2992,7 @@ end
 
 local function getDeveloperConsole()
 	if not myDeveloperConsole then
+		myDeveloperConsole = true -- debounce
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -2992,6 +2992,7 @@ end
 
 local function getDeveloperConsole()
 	if not myDeveloperConsole then
+		myDeveloperConsole = true -- debounce
```
spotco: We're going to use another variable instead of myDeveloperConsole for the debounce.
Your fix will be a part of: https://github.com/ROBLOX/Core-Scripts/pull/540

spotco: I don't know how any of this new github ui works but this change will be a part of #540.

PR #538, opened 2016-09-13
Cleanup, adding 'average' bar

DougBanksPersonal:
Adding new "average" bar to graph to show average of current values.
Adding legend to enlarged graph.
General cleanup: 
- using fixed sizes, dynamic scaling gets too crazy.
- splitting TextPanel class, the maximized and non-maximized versions are different enough it's easier to just split.
- got rid of split notion of "stat button display type" vs "stat type".  Made sense when I was worrying about displaying two stats in a single button.  We are not doing that.  Keeping things saner.


CoreScriptsRoot/Modules/Stats/DecoratedValueLabel.lua
```diff
@@ -0,0 +1,79 @@
...
+  self._label.TextXAlignment = Enum.TextXAlignment.Left
+  self._label.TextYAlignment = Enum.TextYAlignment.Center
+  
+  StatsUtils.StyleTextWidget(self._label)
+  
+  return self
+end
+  
+function DecoratedValueLabelClass:SetZIndex(zIndex)
+  self._frame.ZIndex = zIndex
```
hockeychaos: do any of these overlap? might be a good idea to explicitly set some to a higher zindex if so. By default if zindex is the same we render in child order, but that can always change if mess with parenting items separately.

CoreScriptsRoot/Modules/Stats/StatsAggregator.lua
```diff
@@ -90,7 +90,8 @@ function StatsAggregatorClass:GetAverage()
   for i = 0, self._numSamples-1, 1 do
     retval = retval + self._samples[i]
   end
-  return retval / self._numSamples
+  local average = retval / self._numSamples
```
hockeychaos: this change seem superfluous? for readability?

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
```diff
@@ -20,46 +20,58 @@ local BarGraphClass = require(CoreGuiService.RobloxGui.Modules.Stats.BarGraph)
...
+  StatsUtils.StyleTextWidget(self._midLabel)
+  StatsUtils.StyleTextWidget(self._bottomLabel)
 
   self:_layoutElements()
 
   return self
 end
 
+function StatsAnnotatedGraphClass:SetZIndex(zIndex)
+  self._frame.ZIndex = zIndex
```
hockeychaos: same thing here about index

CoreScriptsRoot/Modules/Stats/DecoratedValueLabel.lua
DougBanksPersonal: They don't overlap with each other.
In general, there's a common pattern of 
Left side has text.
Right side has graph.
We can lay stuff out so each side doesn't have internal overlapping items (nothing on right side runs over anything else on right side).
But the text may go extra long and goosh into the right side, overlapping graph.
Hence the ZIndex functions: once everything is all set up we can set elements in each side to the depth we want.

CoreScriptsRoot/Modules/Stats/StatsAggregator.lua
DougBanksPersonal: Reverted

CoreScriptsRoot/Modules/Stats/StatsAnnotatedGraph.lua
DougBanksPersonal: See above.

PR #540, opened 2016-09-14
SetCore and GetCore DeveloperConsoleVisible.

JacobLocke:
Allows developers to get and set the visibility state of the developer console from Lua.

### DIFF
--- a/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
+++ b/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
@@ -3054,8 +3054,7 @@ StarterGui:RegisterGetCore("DeveloperConsoleVisible", function()
 end)
 StarterGui:RegisterSetCore("DeveloperConsoleVisible", function(visible)
 	if (type(visible) ~= "boolean") then
-		warn("DeveloperConsoleVisible must be given a boolean value.")
-		return
+		error("DeveloperConsoleVisible must be given a boolean value.")
 	end
 
 	if (not myDeveloperConsole) then
--- a/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
+++ b/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
@@ -2990,15 +2990,24 @@ local function onDevConsoleVisibilityChanged(isVisible)
 	end
 end
 
+local getDeveloperConsoleIsCreating = false
 local function getDeveloperConsole()
-	if not myDeveloperConsole then
-		local permissions = DeveloperConsole.GetPermissions()
-		local messagesAndStats = DeveloperConsole.GetMessagesAndStats(permissions)
+	if (not myDeveloperConsole) then
+		if (not getDeveloperConsoleIsCreating) then
+			getDeveloperConsoleIsCreating = true
 
-		myDeveloperConsole = DeveloperConsole.new(RobloxGui, permissions, messagesAndStats)
+			local permissions = DeveloperConsole.GetPermissions()
+			local messagesAndStats = DeveloperConsole.GetMessagesAndStats(permissions)
 
-		if isTenFootInterface then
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -2990,15 +2990,24 @@ local function onDevConsoleVisibilityChanged(isVisible)
 	end
 end
 
+local getDeveloperConsoleIsCreating = false
 local function getDeveloperConsole()
-	if not myDeveloperConsole then
-		local permissions = DeveloperConsole.GetPermissions()
-		local messagesAndStats = DeveloperConsole.GetMessagesAndStats(permissions)
+	if (not myDeveloperConsole) then
+		if (not getDeveloperConsoleIsCreating) then
```
hockeychaos: this is fine, although it seems weird that new must yield at some point... we should look into this in the future.

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -3015,4 +3024,46 @@ function DevConsoleModuleTable:SetVisibility(value)
...
+		--// change.
+		DevConsoleModuleTable:GetVisibility()
+		DevConsoleModuleTable:SetVisibility(creatingVisibleValueToSet)
+
+		creatingLock = false
+	end)
+end
+
+local StarterGui = game:GetService("StarterGui")
+StarterGui:RegisterGetCore("DeveloperConsoleVisible", function()
```
hockeychaos: is it possible for us to define these earlier in the script? I worry about someone trying to call these in a game early on and they don't exist at startup. Could start with an empty function that gets redefined here.

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -3015,4 +3024,46 @@ function DevConsoleModuleTable:SetVisibility(value)
...
+	if (not myDeveloperConsole) then
+		SetCoreConsoleCreation()
+		return creatingVisibleValueToSet;
+	else
+		return DevConsoleModuleTable:GetVisibility()
+	end
+end)
+StarterGui:RegisterSetCore("DeveloperConsoleVisible", function(visible)
+	if (type(visible) ~= "boolean") then
+		warn("DeveloperConsoleVisible must be given a boolean value.")
```
hockeychaos: is this standard practice for Core functions? usually if a lua api gets the wrong arg or is missing one, we throw an error.

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
JacobLocke: The tests I ran had SetCore and GetCore calls immediately after a GetService("StarterGui") in a LocalScript running in StarterPlayerScripts and everything was fine. DeveloperConsoleModule.lua seems to be required at startup.

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
JacobLocke: I chose to do a warning instead of an error because having an error here wouldn't kill the Lua thread that called the SetCore() in the first place, and would show which line in DeveloperConsoleModule.lua the error occurred on instead.

hockeychaos: this won't give you a call stack thought right? If it is standard practice for function calls that we define (whether it is a rbx::reflection or something defined in lua) to throw an error with invalid args, its best to be consistent. Warnings are nice, but they are just warnings. If you are fundamentally using an API wrong, an error seems appropriate.

JacobLocke: Okay. I'll change it to be an error.

PR #541, opened 2016-09-14
VR-242: VR Teleport

kyle-emmerich:
It should be noted that ParabolaAdornment hasn't shipped yet. The second commit in this PR handles that gracefully.

### DIFF
--- a/CoreScriptsRoot/Modules/VR/LaserPointer.lua
+++ b/CoreScriptsRoot/Modules/VR/LaserPointer.lua
@@ -78,8 +78,8 @@ local TELEPORT = {
 	ARC_COLOR_BAD = fromLinearRGB(Color3.fromRGB(253, 68, 72)),
 	ARC_THICKNESS = 0.05,
 
-	PLOP_GOOD = "rbxasset://textures/ui/VR/VR Pointer Disc Blue.png",
-	PLOP_BAD = "rbxasset://textures/ui/VR/VR Pointer Disc Red.png",
+	PLOP_GOOD = "rbxasset://textures/ui/VR/VRPointerDiscBlue.png",
+	PLOP_BAD = "rbxasset://textures/ui/VR/VRPointerDiscRed.png",
 	PLOP_BALL_COLOR_GOOD = BrickColor.new("Bright green"),
 	PLOP_BALL_COLOR_BAD = BrickColor.new("Bright red"),
 	PLOP_BALL_SIZE = 0.5,
@@ -94,7 +94,7 @@ local TELEPORT = {
 
 	MIN_VELOCITY = 10,
 	RANGE_T_EXP = 2,
-	G = 50,
+	G = 50, -- Gravity constant for parabola
 
 	PULSE_DURATION = 0.8,
 	PULSE_PERIOD = 1,
@@ -113,9 +113,11 @@ local TELEPORT = {
 	FADE_IN_DURATION = 0.125,
 
 	CLEAR_AABB_SIZE = Vector3.new(2.5, 4, 2.5),
+	CLEAR_TEST_ENABLED = false,
 
-	SUCCESS_SOUND = "rbxassetid://147722227",
-	FAIL_SOUND = "rbxassetid://138087015",
+	SUCCESS_SOUND = "", --Unused for now
+	FAIL_SOUND = "", --Unused for now
+	SOUNDS_ENABLED = false,
 
 	PATH_RECOMPUTE_DIST_THRESHOLD = 4,
 
@@ -132,7 +134,7 @@ local LASER = {
 
 	MAX_DISTANCE = 100,
 
-	G = 0,
+	G = 0, -- Gravity constant for parabola; in this case we want a laser/straight line
 
 	--Couldn't figure out a good name for this. This is the maximum angle that the parabola's hit point
 	--can be from the laser's hit point when switching to laser pointer mode solely from the parabola hitting
@@ -235,7 +237,7 @@ function LaserPointer.new()
 			Shape = Enum.PartType.Ball,
 			Size = identityVector3 * TELEPORT.PLOP_BALL_SIZE
 		}
-		self.plopAdorn = Utility:Create("ImageHandleAdornment") { --this feels hacky, but no good alternatives for unlit image in 3D that aren't SurfaceGuis w/ ImageLabels
+		self.plopAdorn = Utility:Create("ImageHandleAdornment") {
 			Name = "LaserPointerTeleportPlopAdorn",
 			Parent = self.plopPart,
 			Adornee = self.plopPart,
@@ -251,28 +253,31 @@ function LaserPointer.new()
 			Transparency = 0.5
 		}
 
-		--I want to nuke this part, but for now it's a necessary evil or very bad things happen when you try to teleport
-		--where you don't fit...
-		self.collisionTestPart = Utility:Create("Part") {
-			Name = "LaserPointerTeleportCollisionTester",
-			Size = TELEPORT.CLEAR_AABB_SIZE,
-			Transparency = 1,
-			Anchored = true,
-			CanCollide = true,
-			Parent = workspace.CurrentCamera,
-			CFrame = CFrame.new(1e10, 1e10, 1e10)
-		}
+		if TELEPORT.CLEAR_TEST_ENABLED then
+			--TODO: Before re-enabling CLEAR_TEST_ENABLED, find a better solution
+			self.collisionTestPart = Utility:Create("Part") {
+				Name = "LaserPointerTeleportCollisionTester",
+				Size = TELEPORT.CLEAR_AABB_SIZE,
+				Transparency = 1,
+				Anchored = true,
+				CanCollide = true,
+				Parent = workspace.CurrentCamera,
+				CFrame = CFrame.new(1e10, 1e10, 1e10)
+			}
+		end
 
-		self.teleportSuccessSound = Utility:Create("Sound") {
-			Name = "TeleportSuccessSound",
-			SoundId = TELEPORT.SUCCESS_SOUND,
-			Parent = self.originPart
-		}
-		self.teleportFailSound = Utility:Create("Sound") {
-			Name = "TeleportFailSound",
-			SoundId = TELEPORT.FAIL_SOUND,
-			Parent = self.originPart
-		}
+		if TELEPORT.SOUNDS_ENABLED then
+			self.teleportSuccessSound = Utility:Create("Sound") {
+				Name = "TeleportSuccessSound",
+				SoundId = TELEPORT.SUCCESS_SOUND,
+				Parent = self.originPart
+			}
+			self.teleportFailSound = Utility:Create("Sound") {
+				Name = "TeleportFailSound",
+				SoundId = TELEPORT.FAIL_SOUND,
+				Parent = self.originPart
+			}
+		end
 	end
 
 	do --Event connections and final setup
@@ -549,7 +554,9 @@ do --Action functions
 
 		if teleportValid then
 			--play teleport success sound
-			self.teleportSuccessSound:Play()
+			if TELEPORT.SOUNDS_ENABLED then
+				self.teleportSuccessSound:Play()
+			end
 			self.teleporting = true
 
 			wait(FADE_OUT_DURATION)
@@ -563,7 +570,9 @@ do --Action functions
 			self.teleporting = false
 		else
 			--play teleport failed sound
-			self.teleportFailSound:Play()
+			if TELEPORT.SOUNDS_ENABLED then
+				self.teleportFailSound:Play()
+			end
 		end
 	end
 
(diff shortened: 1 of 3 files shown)


CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,548 @@
...
+
+	FLOOR_OFFSET = 4.5,
+
+	FADE_OUT_DURATION = 0.125,
+	FADE_IN_DURATION = 0.125,
+
+	CLEAR_AABB_SIZE = Vector3.new(2.5, 4, 2.5),
+
+	SUCCESS_SOUND = "rbxassetid://147722227",
+	FAIL_SOUND = "rbxassetid://138087015",
```
kyle-emmerich: Leaving a reminder to myself to change this

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,548 @@
...
+			part.Parent = container
+		end
+	else
+		part.Parent = nil
+	end
+end
+
+--Teleport visual configuration
+local TELEPORT = {
+	ARC_COLOR_GOOD = fromLinearRGB(Color3.fromRGB(0, 162, 255)),
```
SolarCrane: is fromRGB a new function on Color3? Did it go through api committee yet?

CoreScriptsRoot/Modules/VR/LaserPointer.lua
SolarCrane: Let's put this in the included content folder

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,548 @@
...
+local CONTAINER_FOLDER_NAME = "VRCorePanelParts"
+local function setPartInGame(part, inGame)
+	if not part then
+		return
+	end
+	if inGame and not part:IsDescendantOf(game) then
+		local container = workspace.CurrentCamera:FindFirstChild(CONTAINER_FOLDER_NAME)
+		if not container then
+			coroutine.wrap(function()
+				part.Parent = workspace.CurrentCamera:WaitForChild(CONTAINER_FOLDER_NAME)
```
SolarCrane: It doesn't feel great to put this part under the camera, we are putting a lot of core script stuff in the workspace again and again.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,548 @@
...
+			Anchored = true,
+			CanCollide = false,
+			TopSurface = Enum.SurfaceType.SmoothNoOutlines,
+			BottomSurface = Enum.SurfaceType.SmoothNoOutlines,
+			Material = Enum.Material.Neon,
+			BrickColor = TELEPORT.PLOP_BALL_COLOR_GOOD,
+			Shape = Enum.PartType.Ball,
+			Size = identityVector3 * TELEPORT.PLOP_BALL_SIZE
+		}
+		self.plopAdorn = Utility:Create("ImageHandleAdornment") { --this feels hacky, but no good alternatives for unlit image in 3D that aren't SurfaceGuis w/ ImageLabels
```
SolarCrane: What even is an ImageHandleAdornment? I haven't used it before.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,548 @@
...
+	BALL_WAVE_AMPLITUDE = 0.5,
+	BALL_WAVE_START = 0.25,
+	BALL_WAVE_EXP = 0.8,
+
+	FLOOR_OFFSET = 4.5,
+
+	FADE_OUT_DURATION = 0.125,
+	FADE_IN_DURATION = 0.125,
+
+	CLEAR_AABB_SIZE = Vector3.new(2.5, 4, 2.5),
```
SolarCrane: If we are going to use this, we might want to drive this based on your actual character's size

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,548 @@
...
+	local maxBound = minBound + TELEPORT.CLEAR_AABB_SIZE
+
+	local theta = math.rad(90) - math.asin(normal.Y)
+	local slopeOffset = math.sin(theta) * math.sqrt(halfBbSize.X^2 + halfBbSize.Z^2)
+	self.collisionTestPart.CFrame = CFrame.new(point + Vector3.new(0, (TELEPORT.CLEAR_AABB_SIZE.Y / 2) + slopeOffset + 0.1, 0))
+	
+	--Not only does workspace:FindPartsInRegion3 only do AABBs, it also doesn't seem to like slopes.
+	--Using an arbitary part here sounds nasty as an implementation detail, but the concept seems sound.
+	--Getting the part out of the workspace entirely would be a good long-term solution.
+	local foundParts = self.collisionTestPart:GetTouchingParts()
```
SolarCrane: Do you think we can upgrade FindPartsInRegion3 to accept a rotation cframe as well? It seems better than inserting collision objects into the world

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: Yeah, Ethan added it a while ago

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: Hmm, this one we might be able to move out. Arseny has mentioned a desire to remove the requirement of being in the workspace for other things, so we'll eventually be able to move all of these out.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,548 @@
...
+end
+
+--Teleport visual configuration
+local TELEPORT = {
+	ARC_COLOR_GOOD = fromLinearRGB(Color3.fromRGB(0, 162, 255)),
+	ARC_COLOR_BAD = fromLinearRGB(Color3.fromRGB(253, 68, 72)),
+	ARC_THICKNESS = 0.05,
+
+	PLOP_GOOD = "rbxasset://textures/ui/VR/VR Pointer Disc Blue.png",
+	PLOP_BAD = "rbxasset://textures/ui/VR/VR Pointer Disc Red.png",
```
kyle-emmerich: That's how I got 'em! I'll change that though because I agree

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: There are a few of these things added last year by Tyler. I'm not the biggest fan of using them for purely visual purposes, but they work very well for this.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: I can try that. I don't know if we really want to call it FindPartsInRegion3 at that point though; maybe we could add a new method in Workspace that takes a CFrame and Size instead. 

Region3s already have a CFrame but there's no Lua way of setting that right now. We might be able to just add a constructor overload that takes a CFrame and a Size to Region3, but we'll also need to make sure you're not passing rotated Region3s to Terrain methods.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: I doubt we actually want these. I just forgot to remove them. The fail sound is the horn from The Price is Right ;)

That being said, need to talk to product/art about getting actual sounds, because audible confirmation is really nice.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: Since I didn't answer your question, they're an adornment that renders an image on a flat plane a bit like a SurfaceGui, but it has its own CFrame property and renders the image itself rather than through an ImageLabel. Lot less overhead that way.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
SolarCrane: Can we move this one out now rather than later?

CoreScriptsRoot/Modules/VR/LaserPointer.lua
SolarCrane: nifty

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: I have some more stuff coming for this and it's becoming a bit of a moving target. We should hold off on this whole PR.

CoreScriptsRoot/Modules/VR/VRHub.lua
```diff
@@ -2,13 +2,64 @@
...
 
+--VR Setup
+local vrUpdateRenderstepName = HttpService:GenerateGUID(true)
+
+VRHub.LaserPointer = nil
+
+local start, dt = tick(), 1/60
+local function onRenderSteppedLast()
+	local now = tick()
+	dt = now - start
```
SolarCrane: Could we just initialize dt here instead of outside this func? It looks like it isn't really being used anywhere else.

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -689,24 +695,27 @@ end
...
+Subpanel.__index = Subpanel
 function Subpanel.new(parentPanel, guiElement)
-	local self = setmetatable({}, Subpanel_mt)
+	local self = {}
 	self.parentPanel = parentPanel
 	self.guiElement = guiElement
 	self.lastParent = guiElement.Parent
-	self.ancestryConn = nil
-	self.changedConn = nil
+	self.ancestryConn = false
```
kyle-emmerich: Ah, I had done that because I was going to add a __newindex that threw an error in case someone typo'd a member name while setting it. Setting it to nil removes it from the table, so you'd never be able to update it later. I'll change it back.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
SolarCrane: Did we ever fix this?

CoreScriptsRoot/Modules/VR/LaserPointer.lua
SolarCrane: Should update this...

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,933 @@
...
+local LASER = {
+	MODE_ENABLED = true,
+
+	ARC_COLOR_GOOD = TELEPORT.ARC_COLOR_GOOD,
+	ARC_COLOR_BAD = TELEPORT.ARC_COLOR_BAD,
+	ARC_THICKNESS = 0.025,
+
+	MAX_DISTANCE = 100,
+
+	G = 0,
```
SolarCrane: What does G stand for? We should comment this.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,933 @@
...
+			
+			if self.clickToMoveModule then
+				self.clickToMoveModule:Stop()
+			end
+		end
+	end
+
+	function LaserPointer:setTeleportMode(enabled)
+		if enabled == self.teleportMode then return	end
+		if self.tweener then return end
```
SolarCrane: How come we return here is self.tweener?

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,933 @@
...
+		local theta = math.rad(90) - math.asin(normal.Y)
+		local slopeOffset = math.sin(theta) * math.sqrt(halfBbSize.X^2 + halfBbSize.Z^2)
+		self.collisionTestPart.CFrame = CFrame.new(bbPos + Vector3.new(0, (TELEPORT.CLEAR_AABB_SIZE.Y / 2) + slopeOffset + 0.1, 0))
+		
+		--workspace:FindPartsInRegion3 uses AABBs of parts, which means slopes make really big boxes. No good for checking occlusion.
+		--Using an arbitary part here sounds nasty as an implementation detail, but the concept seems sound.
+		--Getting the part out of the workspace entirely would be a good long-term solution.
+		local foundParts = self.collisionTestPart:GetTouchingParts()
+		--send it far away so it doesn't interfere with devs (I really hate doing this)
+		self.collisionTestPart.CFrame = CFrame.new(1e10, 1e10, 1e10)
```
SolarCrane: I think we should rethink of a better solution for this before shippings

CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -0,0 +1,933 @@
...
+		local thickness0, thickness1 = LASER.ARC_THICKNESS, TELEPORT.ARC_THICKNESS
+		local gravity0, gravity1 = LASER.G, TELEPORT.G
+		
+		self.parabola.Thickness = self.transitionAlpha * (thickness1-thickness0) + thickness0
+
+		if self:isHeadMounted() then
+			self.parabola.Thickness = self.parabola.Thickness * HEAD_MOUNT_THICKNESS_MULTIPLIER
+
+			--cast ray from center of camera, then render laser going from offset point to hit point
+			local originCFrame = cameraSpace * UserInputService:GetUserCFrame(Enum.UserCFrame.Head)
```
SolarCrane: Are we moving this GetUserCFrame to the VRService eventually too?

kyle-emmerich: Yeah, that's the plan. Once it ships for all platforms I'll come back and change all of them over. The behavior will remain the same.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: It's to avoid switching during a transition. Maybe tweener needs a better name.

CoreScriptsRoot/Modules/VR/LaserPointer.lua
kyle-emmerich: Just now 👍

PR #542, opened 2016-09-14
Load VR AppShell code

SolarCrane:


CoreScriptsRoot/StarterScript.lua
```diff
@@ -66,3 +66,30 @@ if useVRKeyboard then
 	require(RobloxGui.Modules.VR.VirtualKeyboard)
 end
 
+
+-- Boot up the VR App Shell
+if UserSettings().GameSettings:InStudioMode() then
```
hockeychaos: poor starter script...

CoreScriptsRoot/StarterScript.lua
```diff
@@ -66,3 +66,30 @@ if useVRKeyboard then
...
+				local appHomeModule = modulesFolder:FindFirstChild('Shell') and modulesFolder:FindFirstChild('Shell'):FindFirstChild('AppHome')
+				if shellInVR and appHomeModule then
+					require(appHomeModule)
+				end
+			end
+		end
+	end
+
+	UserInputService.Changed:connect(onVREnabled)
+	spawn(function()
```
hockeychaos: shouldn't we do this before connecting to event?

CoreScriptsRoot/StarterScript.lua
SolarCrane: If it would be preferable, we could refactor this following codeblock into its own module. That way we don't have complicated logic in the StarterScript

hockeychaos: meh, lets just leave this here for now and move it when we do the thing

PR #544, opened 2016-09-14
Add 'target' logic

DougBanksPersonal:
Show bar of target value for each stat.
Color bars and current value widget according to whether we're at or over target.


CoreScriptsRoot/Modules/Stats/BarGraph.lua
```diff
@@ -19,31 +19,49 @@ local StatsUtils = require(CoreGuiService.RobloxGui.Modules.Stats.StatsUtils)
...
+  self._lineFrame = Instance.new("Frame")
+  self._lineFrame.Name = "PS_LineFrame"
+  self._lineFrame.BackgroundTransparency = 1.0
+  
+  self._showExtras = showExtras
   
   
   self._values = {}
   self._average = 0
+  self._target = 0
```
hockeychaos: could you briefly explain what target accomplishes? Having some trouble following thru in the code... is it for showing an ideal performance standard?

DougBanksPersonal:
Yes.  I will add comments.

DougBanksPersonal:
Added.  Please take a look.

PR #545, opened 2016-09-15
Moving debounce to a lower level

einsteinK:
Moving the debounce from *getDeveloperConsole()* to *DeveloperConsole.GetPermissions()*
(Fixing GetPermissions() also fixes getDeveloperConsole() as the first one was the actual culprit)

After reviewing some stuff, it seems that fixing the culprit (GetPermissions() yielding) seems better.
Having GetPermissions() be a true singleton, instead of singleton-after-it-loaded also fixes getDeveloperConsole().

*(I should've done this in my pull request yesterday, but oh well)*

### DIFF
--- a/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
+++ b/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
@@ -2719,12 +2719,16 @@ end
 -- Permissions --
 -----------------
 do
-	local permissions;
+	local permissionsLoading, permissions = false;
 	function DeveloperConsole.GetPermissions()
+		while permissionsLoading do wait() end
+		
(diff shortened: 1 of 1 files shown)


garnold24:
I was testing this out and noticed that although it fixes the developer console breaking and not showing up, it now duplicates the console if you spam it before you're fully loaded in.

einsteinK:
Hmm that didn't happen when I tested it, but it seems like it could lead to that.
Just adding a debounce to the permissions is probably a better idea.
(Not that it matters much currently because it only gets called by getDeveloperConsole(), for now)

spotco:
Included in 
https://github.com/ROBLOX/Core-Scripts/pull/552

PR #550, opened 2016-09-19
Update Sound.rbxmx

spotco:
http://devforum.roblox.com/t/optimized-and-improved-character-sound-script/28408/15

Was playing disaster survival and noticing that I was hearing too many footsteps for the amount of people around me. Decreased max distance and increased falloff.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
@@ -23,7 +23,11 @@ function CreateNewSound(name, id, looped, pitch, parent)
 	sound.archivable = false
 	sound.Parent = parent
 	sound.Pitch = pitch
-	sound.Looped = looped	
+	sound.Looped = looped
+	
+	sound.MinDistance = 5
+	sound.MaxDistance = 150
+	sound.Volume = 0.65
 	
 	return sound
 end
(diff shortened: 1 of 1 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterCharacterScripts/Sound.rbxmx
```diff
@@ -83,7 +85,15 @@ local Sounds = {}
 do
 	local Figure = script.Parent.Parent
 	Head = Figure:WaitForChild("Head")
-	Humanoid = Figure:WaitForChild("Humanoid")
```
hockeychaos: I understand we want a type of humanoid and thats why you changed this, but why just for humanoid? Head doesn't wait for an actual head

hockeychaos: also does this mean we need a better way to wait for child, with additional parameters or something?

spotco: see: http://devforum.roblox.com/t/optimized-and-improved-character-sound-script/28408/13

hockeychaos: doesn't this still bode the same for head? what if someone renames head? or if the object with the name "Head" is no longer the head? Do we need another WaitForChild? At the least shouldn't we wait for both of these in the same way?

spotco: Can't wait for head typechecked because head is just a part, so that's something we've gotta keep.

hockeychaos: yeah, but you could add something named head that isn't a part right? Would it make sense to make a function here that looks for a type and optionally a name as well?

spotco: sound has to be child of a part (Part object) in order to be 3D. this can be changed to just a general "IHasLocation" but wouldn't be any outward differences.

PR #552, opened 2016-09-20
CLI-13157 Dev Console broken when opened too early

garnold24:
einsteinK seemed to have fixed this with his PR, so i included that in this one, but also fixed an issue which arose that involved the console duplicating itself.

### DIFF
--- a/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
+++ b/CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
@@ -2995,8 +2995,10 @@ local function onDevConsoleVisibilityChanged(isVisible)
 	end
 end
 
+local devConsoleCreating = false
 local function getDeveloperConsole()
-	if (not myDeveloperConsole) then
+	if (not myDeveloperConsole and not devConsoleCreating) then
+		devConsoleCreating = true
 		local permissions = DeveloperConsole.GetPermissions()
 		local messagesAndStats = DeveloperConsole.GetMessagesAndStats(permissions)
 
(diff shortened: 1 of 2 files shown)


CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -2719,12 +2719,16 @@ end
 -- Permissions --
 -----------------
 do
-	local permissions;
+	local permissionsLoading, permissions = false;
```
hockeychaos: if I'm not mistaken, I don't think permissionsLoading is initialized here, although it starts as nil, which is very similar to false.

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -2719,12 +2719,16 @@ end
...
 	function DeveloperConsole.GetPermissions()
+		while permissionsLoading do wait() end
```
hockeychaos: seems the only reason this is necessary is due to HttpRbxApiService. Would it be possible just to spawn a thread for this call so we don't need to debounce?

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -2990,38 +2995,37 @@ local function onDevConsoleVisibilityChanged(isVisible)
 	end
 end
 
-local getDeveloperConsoleIsCreating = false
+local devConsoleCreating = false
```
hockeychaos: I believe this also stems from the http call, if we can isolate that I think we can make this change a lot smaller

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
garnold24: i think permissionsLoading gets initialized to false, while permissions is nil.

CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
garnold24: Even if we spawn a new thread, I'm pretty sure we'd need to still keep the debounce, as everytime you attempt to open the console, it's going to check permissions. If we remove the debounce it's going to keep calling all of that api stuff everytime it wants to open until it goes through and creates permissions, which i don't think is what we want to have happen.

hockeychaos: hmm alright, I guess permissions could change in the middle of a session, but seems unnecessary to do this more than once, as it seems to complicate the code quite a bit. I'm ok with this change for now, but we should consider moving the permission check to script start up

PR #554, opened 2016-09-21
CLISTUDIO-9580: SpawnLocation doesn't orient correctly in first person

tberg-roblox:
Setting CameraMode to LockFirstPerson will cause the ZoomCamera call to
zoom to 0. This causes the character to move to face camera. We want to
orient the camera before we zoom

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
@@ -1115,7 +1115,7 @@ local function CreateCamera()
 		end)
 
 		local function OnCharacterAdded(newCharacter)
-			this:ZoomCamera(12.5)
+			
 			local humanoid = findPlayerHumanoid(player)
 			local start = tick()
 			while tick() - start < 0.3 and (humanoid == nil or humanoid.Torso == nil) do
@@ -1139,8 +1139,10 @@ local function CreateCamera()
 					this.LastCameraTransform = nil
 				end
 			end
-			wait()
+			
 			setLookBehindCharacter()
+			wait()
+			this:ZoomCamera(12.5)
 		end
 
 		player.CharacterAdded:connect(function(character)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -1139,8 +1139,10 @@ local function CreateCamera()
 					this.LastCameraTransform = nil
 				end
 			end
-			wait()
+			
 			setLookBehindCharacter()
+			wait()
```
hockeychaos: is this wait actually necessary? maybe @SolarCrane can shed some light?

SolarCrane: I think it was necessary because at the time of adding the character the engine had not actually rotated your character yet and you would spawn looking in the wrong direction.

hockeychaos: ah, so probably still necessary?

hockeychaos: could we add a comment about that so it doesn't keep getting brought up?

hockeychaos: @tberg-roblox pls

SolarCrane: I think we can pull this in but to be safe we should be a user fast flag on this one

hockeychaos: I think its safe too, but I'd like to put a comment as to why this is necessary

hockeychaos: @tberg-roblox bump

hockeychaos: @tberg-roblox bump 2

PR #555, opened 2016-09-22
Final Polish on Quick Profiler

DougBanksPersonal:
- use "blue underline" from settings menu to indicate selected stat.
- remove 'mid' value on y-axis of graph.
- use image with rounded corners as background of graph.
- move the chat window down & out of the way when stats are present.
- be consistent with whitespace vs tabs (get rid of tabs).


CoreScriptsRoot/Modules/StyleWidgets.lua
```diff
@@ -0,0 +1,40 @@
...
+----------- CONSTANTS --------------
+StyleWidgets.TabSelectionHeight = 6
+
+function StyleWidgets.MakeTabSelectionWidget(parent)
+	local tabSelection = utility:Create'ImageLabel'
+	{
+		Name = "TabSelection",
+		Image = "rbxasset://textures/ui/Settings/MenuBarAssets/MenuSelection.png",
+		ScaleType = Enum.ScaleType.Slice,
+		SliceCenter = Rect.new(3,1,4,5),
```
SolarCrane: Shouldn't the slice on this asset be: 2,2 5,4 ?

DougBanksPersonal: I just did a cut and paste from 
CoreScriptsRoot/Modules/Settings/SettingsPageFactory.lua

See line 96 in the old file.

So maybe it should be but that could be an independent change.

CoreScriptsRoot/StarterScript.lua
```diff
@@ -20,7 +20,7 @@ local luaControlsSuccess, luaControlsFlagValue = pcall(function() return setting
 local vrKeyboardSuccess, vrKeyboardFlagValue = pcall(function() return settings():GetFFlag("UseVRKeyboardInLua") end)
 local useVRKeyboard = (vrKeyboardSuccess and vrKeyboardFlagValue == true)
 
--- MainBotChatScript (the Lua part of Dialogs)
+-- MainBotChatScript (the Lua part of Dialogs)
```
DougBanksPersonal: Done.

DougBanksPersonal:
Tabs to spaces: 
That was deliberate.  Both files used a mix of tabs and spaces, and the diffs were a mess.
Why undo it?  We could go back and forth on whether tabs or spaces are better but having a mix is just confusing.

At this point, to restore the file to what it was would basically mean reverting all of my changes to those files and then adding back in the non-space changes by hand, which feels like a lot of work to ultimately leave the files in a worse condition (mixed) than they're currently in (consistent).

Shiny chat changes: I will talk to him.  I am hoping this could go live ASAP so unless his business is ready to go like today or Monday I'd like to go ahead with this change.

PR #557, opened 2016-09-22
Fixing some crashes with moving xbox files

SolarCrane:


CoreScriptsRoot/CoreScripts/NotificationScript2.lua
```diff
@@ -825,13 +825,21 @@ end
 local UserInputService = game:GetService('UserInputService')
 local Platform = UserInputService:GetPlatform()
 local Modules = RobloxGui:FindFirstChild('Modules')
+local CSMModule = Modules:FindFirstChild('ControllerStateManager')
+if Modules and not CSMModule then
+	local ShellModules = Modules:FindFirstChild('Shell')
```
hockeychaos: will we be able to remove these ifs once the shell conversion process is over?

SolarCrane: yes

PR #558, opened 2016-09-23
CLI-3017 Mobile Jump Button Should Disappear If Jumping Disabled

garnold24:
This edit makes it so that the jump button disappears if the character's humanoid's jump power is set to 0. If it goes higher, then the button gets reenabled.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/TouchJump.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/TouchJump.rbxmx
@@ -35,7 +35,9 @@ local TOUCH_CONTROL_SHEET = "rbxasset://textures/ui/TouchControlsSheet.png"
 
 --[[ Public API ]]--
 function TouchJump:Enable()
-	JumpButton.Visible = true
+	if humanoid.JumpPower > 0 and not JumpButton.Visible then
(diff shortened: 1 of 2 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/TouchJump.rbxmx
```diff
@@ -97,6 +100,18 @@ function TouchJump:Create(parentFrame)
 		end
 	end)
 	
+	if humanoid then
+		
+		humanoid.Changed:connect(function(prop)
```
hockeychaos: I think for this we would want to store the connection and disconnect on disable, since humanoid can change. We should probably also listen to player humanoid changing? would make this more robust + change in MasterControl unnecessary.

garnold24: Oh yeah, I completely forgot PlayerScripts don't reset on respawn. As it is now, when you respawn, humanoid will still be pointing to the old humanoid.

garnold24: My reasoning for using MasterControl was to avoid writing a similar function to get the humanoid in TouchJump, which I feel would have been redundant.

PR #561, opened 2016-09-27
Add variable wait for leave game

hockeychaos:
need leave game to happen sometime after the button is pressed, as slower devices need to update the button color to indicate to the user the button was pressed


kyle-emmerich:
you're a very productive man

PR #567, opened 2016-09-30
New Chat Message Creator modules.

ConorGriffin37:
Also fix blue textbox bug.


CoreScriptsRoot/Modules/Server/ClientChat/ChatMain.lua
```diff
@@ -548,7 +548,16 @@ local function HandleChannelJoined(channel, welcomeMessage, messageLog)
 		end
 
 		if (welcomeMessage ~= "") then
-			channelObj:AddMessageToChannel(welcomeMessage, "WelcomeMessage")
+			welcomeMessageObject = {
```
spotco: Can we unify these keys somewhere in some util class?
welcomeMessageObject = {
   [MessageObject.KEY_ID] = -1,
   [MessageObject.KEY_FROM_SPEAKER_NAME] = nil,
   ...
}

CoreScriptsRoot/Modules/Server/ClientChat/DefaultClientChatModules/MessageCreatorModules/ChannelEchoMessage.lua
```diff
@@ -0,0 +1,46 @@
+local MESSAGE_TYPE = "ChannelEchoMessage"
```
spotco: Any chance you can write out an API for these modules? (What different hooks are there for the different parts?) 
(Once this is all set in stone, that is).

spotco:
Three tasks:
1. Make the util.KEY_CREATOR_FUNCTION return an object with keys
2. Add a header to all the default MessageCreatorModules (we're probably gonna be adding a bunch more hooks. think on click, etc)
3. Headers for all the default MessageCreatorModules.

Good once we have that.

PR #569, opened 2016-10-05
Fix sliders in settings menu

kyle-emmerich:
Apparently no one uses the super nice dragging feature of sliders because I've seen exactly zero reports of this bug...


CoreScriptsRoot/Modules/Settings/Utility.lua
```diff
@@ -1698,7 +1698,7 @@ local function CreateNewSlider(numOfSteps, startStep, minStep)
 
 	local function mouseUpFunc(inputObject)
 		if not interactable then return end
-		if not isActivateEvent() then return end
+		if not isActivateEvent(inputObject) then return end
```
hockeychaos: has this fix been testing with all types of input? keyboard, mouse, touch and controller?

PR #572, opened 2016-10-06
Add scaling for camera height

DarraghGriffin:


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -126,7 +127,7 @@ local function CreateCamera()
 	this.RotateInput = Vector2.new()
 
 	this.lastSubject = nil
-	this.lastSubjectPosition = Vector3.new(0, 0, 0)
+	this.lastSubjectPosition = Vector3.new(0, 5, 0)
```
hockeychaos: whats up with the arbitrary 5 here?

DarraghGriffin: This is just so the camera no longer starts in the ground when the character hasn't loaded yet. Maybe we should solve this in another way though as some games have levels with different ground levels or spawns in different positions.

PR #583, opened 2016-10-11
Add PascalCase aliases to Signal methods

GigsD4X:
Fix to align Signal method names consistently with recent deprecation of :connect and :disconnect in favor of :Connect and :Disconnect (http://devforum.roblox.com/t/roblox-official-changelog/25964/18)


hockeychaos:
While libraries are still used in some places, these are planned for deprecation.

PR #586, opened 2016-10-12
Remove old flag that was never enabled

DarraghGriffin:


DarraghGriffin:
This pull request is redundant, see #587

PR #587, opened 2016-10-12
Shortcut key to toggle quick profiler

DougBanksPersonal:
Whitespace changes: tabs to spaces.


DarraghGriffin:
I don't think we should convert the tabs to spaces in the repository, maybe one of the maintainers (cc @hockeychaos, @SolarCrane @jmargh) will have thoughts on it though.

hockeychaos:
I really don't care, but if I was to choose I'd go tabs... see super helpful video here: https://youtu.be/SsoOG6ZeyUI?t=37 @SolarCrane probably has a pref I think though

PR #600, opened 2016-10-19
Change LaserPointer actions from BindCoreAction to BindAction so deve…

kyle-emmerich:
…lopers can bind over it and use the A button for game actions. Also flag the ClassicCamera changes, forgot to commit this yesterday.

### DIFF
--- a/CoreScriptsRoot/Modules/VR/LaserPointer.lua
+++ b/CoreScriptsRoot/Modules/VR/LaserPointer.lua
@@ -416,17 +416,17 @@ do --Input handlers - can we eventually move this stuff to PlayerScripts? CoreSc
 
 	function LaserPointer:setTriggerActionEnabled(enabled)
 		if enabled then
-			ContextActionService:BindCoreAction("TeleportTriggerImpl", function(...) self:onTeleportTriggerAction(...) end, false, self.inputUserCFrame == Enum.UserCFrame.RightHand and Enum.KeyCode.ButtonA or Enum.KeyCode.ButtonX)
+			ContextActionService:BindAction("TeleportTriggerImpl", function(...) self:onTeleportTriggerAction(...) end, false, self.inputUserCFrame == Enum.UserCFrame.RightHand and Enum.KeyCode.ButtonA or Enum.KeyCode.ButtonX)
 		else
-			ContextActionService:UnbindCoreAction("TeleportTriggerImpl")
+			ContextActionService:UnbindAction("TeleportTriggerImpl")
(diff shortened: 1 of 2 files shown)


CoreScriptsRoot/Modules/VR/LaserPointer.lua
```diff
@@ -416,17 +416,17 @@ do --Input handlers - can we eventually move this stuff to PlayerScripts? CoreSc
 
 	function LaserPointer:setTriggerActionEnabled(enabled)
 		if enabled then
-			ContextActionService:BindCoreAction("TeleportTriggerImpl", function(...) self:onTeleportTriggerAction(...) end, false, self.inputUserCFrame == Enum.UserCFrame.RightHand and Enum.KeyCode.ButtonA or Enum.KeyCode.ButtonX)
+			ContextActionService:BindAction("TeleportTriggerImpl", function(...) self:onTeleportTriggerAction(...) end, false, self.inputUserCFrame == Enum.UserCFrame.RightHand and Enum.KeyCode.ButtonA or Enum.KeyCode.ButtonX)
```
kyle-emmerich: The justification for using BindAction here, in my mind, is that this code should really be under ControlScript anyway, but we can't have it there because ParabolaAdornment is limited to CoreScripts right now. I'm making a note to move this module if ParabolaAdornment gets that limitation lifted.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera/ClassicCamera.rbxmx
```diff
@@ -14,6 +14,14 @@ if not VRServiceExists or not VRService then
 	VRService = UserInputService
 end
 
+local VRCameraUpdate = false
+local success, msg = pcall(function()
+	VRCameraUpdate = UserSettings():IsUserFeatureEnabled("VRCameraUpdateLua")
```
SolarCrane: The `IsUserFeatureEnabled` is generally not used because they have an undesirable effect if the dev copies the camera scripts and then we remove this flag.
Is this feature something that we need to be worried about turning on and off?

kyle-emmerich: In that case, probably not. I'll remove it from this PR.

PR #605, opened 2016-10-21
VR-344 VR app panel bug fixes

SolarCrane:


kyle-emmerich:
:+1: I don't have write access but I reviewed it and it looks good.

PR #608, opened 2016-10-24
Ctrl-F7, instead of just f7, to make PerformanceStats show up.

DougBanksPersonal:
Use Ctrl key as modifier on F7 to toggle Performance Stats visibility.


CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -554,9 +554,15 @@ local function CreateSettingsHub()
...
   end
 
-  local function toggleQuickProfiler(actionName, inputState, inputObject) 
-      -- F7
-      if actionName ==QUICK_PROFILER_ACTION_NAME then
+  local function toggleQuickProfilerFromHotkey(actionName, inputState, inputObject) 
+    -- Make sure it's Ctrl-F7.
+    -- NOTE: This will only work if FFlagDontSwallowInputForStudioShortcuts is True.
+    -- Otherwise, we never get the "Begin" input state when Ctrl key is down.
+    if (not UserInputService:IsKeyDown(Enum.KeyCode.LeftControl)) then
```
hockeychaos: do we care about right control?

SolarCrane: Right control is out of control!

DougBanksPersonal:
I added a case to check for Right Control also.

PR #612, opened 2016-10-27
Use SetTextFromInput to update text from the VR Keyboard

jovannic:
Trying to make sure that any place TextBox text is updated with user input outside of C++ that it goes through this (new) function.


CoreScriptsRoot/Modules/VR/VirtualKeyboard.lua
```diff
@@ -968,7 +971,15 @@ local function ConstructKeyboardUI(keyboardLayoutDefinitions)
 	end
 
 	local function UpdateTextEntryFieldText(newText)
-		textEntryField.Text = newText
+		local textSet = false
+		if useSetTextFromInput then
+			local success, textSet = pcall(function() return textEntryField:SetTextFromInput(newText) end)
+		end
+
+		if not textSet then
```
hockeychaos: should this be if not success or not textSet ?

jovannic:
Fixed.

PR #615, opened 2016-10-31
Set parent last for instances created with Utility.Create

DarraghGriffin:
See
http://devforum.roblox.com/t/psa-dont-use-instance-new-with-parent-argument/30296

### DIFF
--- a/CoreScriptsRoot/Libraries/RbxUtility.lua
+++ b/CoreScriptsRoot/Libraries/RbxUtility.lua
@@ -927,6 +927,7 @@ local function Create_PrivImpl(objectType)
 
 		--make the object to mutate
 		local obj = Instance.new(objectType)
+		local parent = nil
 
 		--stored constructor function to be called after other initialization
 		local ctor = nil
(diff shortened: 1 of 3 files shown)


CoreScriptsRoot/Libraries/RbxUtility.lua
```diff
@@ -975,6 +982,10 @@ local function Create_PrivImpl(objectType)
 		if ctor then
 			ctor(obj)
 		end
+		
+		if parent then
```
kyle-emmerich: FWIW, you don't actually need this check here, but oh well

kyle-emmerich:
https://github.com/ROBLOX/Core-Scripts/blob/master/CoreScriptsRoot/Libraries/RbxUtility.lua#L916

Should also update this one

CoreScriptsRoot/Libraries/RbxUtility.lua
```diff
@@ -927,14 +927,21 @@ local function Create_PrivImpl(objectType)
...
+		local parent = nil
 
 		--stored constructor function to be called after other initialization
 		local ctor = nil
 
 		for k, v in pairs(dat) do
 			--add property
 			if type(k) == 'string' then
-				obj[k] = v
+				if k == 'Parent' then
```
SolarCrane: Should we FFlag this change? While it seems to be better all around there could be some unintended consequences.

DarraghGriffin: I'll flag it to be safe

PR #626, opened 2016-11-06
Fix redundant badge notifications

foreverpower:
Fix bug where multiple redundant notifications could be shown for the same badge if BadgeAwared is called multiple times in quick succession. This tends to happen with badge-giving bricks, among other things.


foreverpower:
That should fix it.

PR #627, opened 2016-11-06
Kick players when they leave

PhoenixShay:
Kick a player when their player is removed, to prevent an exploiter from placing themselves into nil to avoid being indexed.

### DIFF
--- a/CoreScriptsRoot/ServerStarterScript.lua
+++ b/CoreScriptsRoot/ServerStarterScript.lua
@@ -57,6 +57,7 @@ game:GetService("Players").PlayerRemoving:connect(function(player)
 			end
 		end
 	end
+	player:Kick()
 end)
 
 local success, retVal = pcall(function() return game:GetService("Chat"):GetShouldUseLuaChat() end)


DarraghGriffin:
Re-parenting your player object should already cause you to be disconnected (and if not this is a bug we should fix).

Regardless I don't think this is the right fix for this issue, it might have unintended consequences for players who are not exploiting.

foreverpower:
Currently, reparenting the player object doesn't cause the player to be disconnected, and there is also not a way to kick them even if you have a reference to their player object, as player:Kick() will not function when their parent is nil.

PhoenixShay:
Foreverpower, going into nil doesn't change the fact that it's a player instance, and all player instances can be kicked, no matter where they are. As for any unintended consequences that may happen, unless the PlayerRemoving event fires incorrectly (which is a ROBLOX fault and is a problem nonetheless), then you don't have to worry about it.

Yes, exploiters often parent themselves to nil, and they still are connected to the game server and able to manipulate the world (though their character goes into nil with them, obviously).

foreverpower:
The method can still be called, yes, but it does not function. Regardless, I don't think this is the correct way to go about doing this, although this is a problem worth solving.

kyle-emmerich:
Definitely should not do this. Kick() is not intended for this purpose and may have side effects such as event reporting for analytics.

PR #628, opened 2016-11-07
	Make setting ResetButtonCallback to true restore default reset

DarraghGriffin:


CoreScriptsRoot/Modules/Settings/Pages/ResetCharacter.lua
```diff
@@ -86,14 +86,15 @@ local function Initialize()
...
-	this.ResetBindable = Instance.new("BindableEvent")
-	this.ResetBindable.Event:connect(resetCharFunc)
+	this.ResetBindable = true
 	
 	local onResetFunction = function()
 		if this.HubRef then
 			this.HubRef:SetVisibility(false, true)
 		end
-		if this.ResetBindable then
+		if this.ResetBindable == true then
```
hockeychaos: why not just have ResetBindable set to nil and do this behavior?

PR #633, opened 2016-11-12
New chat fix name color

foreverpower:
Update the user's name color upon switching teams


ConorGriffin37:
This is a good spot, name colors should definitely update when players change teams. This fix looks good to me.

PR #639, opened 2016-11-15
Fix redundant badge notification.

foreverpower:
Fix badge notifications being displayed twice if :AwardBadge is called in quick succession.


DarraghGriffin:
This pull request looks good to me. I think we should merge it.

PR #654, opened 2016-12-09
Bubblechat use new chat events

foreverpower:
Bubblechat switch to new chat events


ConorGriffin37:
Thanks for your pull request. We are already working on this internally and will be moving the bubble chat script out of the corescripts and into the new chat in january.

PR #656, opened 2016-12-10
New chat various fixes

foreverpower:
Fix mute case sensitivity, add guest check to ChatMessageValidator, and correct misspelling in UnknownMessage.lua


ConorGriffin37:
This fix is good, I don't know how I didn't realize to compare speaker names in lower case :(
Preventing guests from chatting on the server as well as the client is also a good idea.

PR #661, opened 2016-12-21
Fix SetCore("ResetButtonCallback") on mobile

DarraghGriffin:
The issue was that the initialization was contained in a code path that
wan't being hit on Mobile devices.

Also switch this script back to using tab indents instead of spaces.
Reasoning: Nearly every script in the CoreScript GitHub uses tabs for
indentation, we should be consistent with this.

### DIFF
--- a/CoreScriptsRoot/Modules/Server/ClientChat/ChatMain.lua
+++ b/CoreScriptsRoot/Modules/Server/ClientChat/ChatMain.lua
@@ -145,7 +145,9 @@ spawn(function()
 		local tickDelta = currentTick - lastTick
 		local dtScale = CurveUtil:DeltaTimeToTimescale(tickDelta)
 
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -49,1155 +49,1177 @@ local resetButtonFlagSuccess, resetButtonFlagValue = pcall(function() return set
...
-            if GuiService.SelectedCoreObject == nil then
-              this.Pages.CurrentPage:SelectARow()
-            end
-          end
-        end
-      end
-    end)
-
-  return this
+	local this = {}
```
hockeychaos: is this diff old? It would be nice to not change all the whitespace

PR #662, opened 2016-12-21
New Chat fix background transparency bug

ConorGriffin37:


CoreScriptsRoot/Modules/Server/ClientChat/ChatMain.lua
```diff
@@ -140,6 +140,7 @@ spawn(function()
 
 	local updateWaitTime = 1.0 / animationFps
 	local lastTick = tick()
+	wait(updateWaitTime)
```
spotco: if currentTime == lastTime then
continue

/\ do this instead /\

PR #673, opened 2017-01-06
Move BubbleChat outside of the corescripts.

ConorGriffin37:
Make bubble chat part of the new chat outside of the corescripts. This makes it easy for developers to edit the look of Bubble chat in their games :)

### DIFF
--- a/CoreScriptsRoot/Modules/NewChat.lua
+++ b/CoreScriptsRoot/Modules/NewChat.lua
@@ -133,7 +133,7 @@ do
 			return not (BubbleChatEnabled or ClassicChatEnabled)
 		end
 
-		function SetInitalChatTypes(chatTypesTable)
+		function SetInitialChatTypes(chatTypesTable)
 			if ChatTypesSet then
 				return
 			end
@@ -224,7 +224,7 @@ do
 					end
 
 					if type(chatWindowCollection.ChatTypes) == "table" then
-						SetInitalChatTypes(chatWindowCollection.ChatTypes)
+						SetInitialChatTypes(chatWindowCollection.ChatTypes)
 					end
 
 					eventConnections = {}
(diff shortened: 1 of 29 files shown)


CoreScriptsRoot/Modules/NewChat.lua
```diff
@@ -116,11 +121,48 @@ do
...
+
+		function moduleApiTable:IsBubbleChatOnly()
+			return BubbleChatEnabled and not ClassicChatEnabled
+		end
+
+		function moduleApiTable:IsDisabled()
+			return not (BubbleChatEnabled or ClassicChatEnabled)
+		end
+
+		function SetInitalChatTypes(chatTypesTable)
```
spotco: initial

spotco:
make example

foreverpower:
https://github.com/ROBLOX/Core-Scripts/pull/673/files#diff-a219ffb2f286125651b431ac2c44c5b6R550

This line looks incomplete.

ConorGriffin37:
@foreverpower It's completed on the next 2 lines. Github just diffs it weirdly.

CoreScriptsRoot/Modules/Server/ClientChat/BubbleChat/BubbleChat.lua
```diff
@@ -0,0 +1,693 @@
...
+local OnMessageDoneFiltering = ChatEvents:WaitForChild("OnMessageDoneFiltering")
+local OnNewMessage = ChatEvents:WaitForChild("OnNewMessage")
+
+OnNewMessage.OnClientEvent:connect(function(messageData, channelName)
+	local sender = findPlayer(messageData.FromSpeaker)
+	if not sender then
+		return
+	end
+
+	if not messageData.IsFiltered then
```
foreverpower: This should be changed to:

`if not messageData.IsFiltered or messageData.FromSpeaker == LocalPlayer.Name then`

because IsFiltered is automatically set to true for all messages sent by the local player, despite not actually being filtered. This would cause all messages from the local player to be duplicated in the event that ShowOwnFilteredMessage is true.

ConorGriffin37: Nice spot. I have added a fix for this. I tested this but I actually wasn't reading ShowUserOwnFilteredMessage correctly, so it was always false.
https://github.com/ROBLOX/Core-Scripts/pull/673/commits/95fcb45dc5b5c4b4d424c602f503a9be3f1855f0

PR #674, opened 2017-01-08
TargetChannel should initialize as a blank string.

CloneTrooper1019:
The ClassMaker module thinks that TargetChannel isn't a valid property before it gets set, which results in the chat getting spammed with errors when typing into the chat bar before it gets set. Setting it to a blank string fixes this problem.

This issue only happens if ShowChannelsBar is set to true in the ChatSettings.


ConorGriffin37:
I have wanted to kill the class maker module for awhile now. It leads to us using a bunch of rawgets and sets all over the place and bugs like this one.

spotco:
No longer necessary because classmaker has been removed. Thanks!

PR #677, opened 2017-01-10
VehicleHud bug fix

foreverpower:
Fix VehicleHud not hiding when Player:LoadCharacter() is called while the player is in a VehicleSeat. Also switch to Heartbeat and clean up old code.


hockeychaos:
@jmargh or @DarraghGriffin can you guys check this out?

DarraghGriffin:
@hockeychaos This looks good to me, I think we should merge it.

jmargh:
@hockeychaos 👍

PR #679, opened 2017-01-11
Fix following icon

foreverpower:
Fix all following status icons being overridden when a new player joins the game. This happens because the following status of a new player is sent in the same format as the initial data. Thus, the script incorrectly attempts to set all existing players' icons to nil because no data is sent for already existing players.


hockeychaos:
@jmargh maybe you should take a look?

PR #685, opened 2017-01-17
Prompt Send Friend Request and GetCore player events.

ConorGriffin37:


CoreScriptsRoot/CoreScripts/FriendPlayerPrompt.lua
```diff
@@ -0,0 +1,180 @@
...
+	local function promptCompletedCallback(clickedConfirm)
+		if clickedConfirm then
+			local successfullyUnfriended = UnFriendPlayer(playerToUnfriend)
+			if not successfullyUnfriended then
+				while PromptCreator:IsCurrentlyPrompting() do
+					wait()
+				end
+				PromptCreator:CreatePrompt({
+					WindowTitle = "Error Unfriending Player",
+					MainText = string.format("An error occured while unfriending %s. Please try again later.", playerToUnfriend.Name),
```
foreverpower: Occured is misspelled. Should be occurred.

CoreScriptsRoot/CoreScripts/FriendPlayerPrompt.lua
```diff
@@ -0,0 +1,180 @@
...
+function AtFriendLimit()
+	local friendCount = PlayerDropDownModule:GetFriendCountAsync(LocalPlayer)
+	if friendCount >= PlayerDropDownModule:MaxFriendCount() then
+		return true
+	end
+	return false
+end
+
+function DoPromptRequestFriendPlayer(playerToFriend)
+	if LocalPlayer:IsFriendsWith(playerToFriend.UserId) then
```
foreverpower: LocalPlayer:GetFriendStatus should be used because :IsFriendsWith caches and is not based on the in-game value.

CoreScriptsRoot/CoreScripts/FriendPlayerPrompt.lua
```diff
@@ -0,0 +1,180 @@
...
+		return
+	end
+	local function promptCompletedCallback(clickedConfirm)
+		if clickedConfirm then
+			if AtFriendLimit() then
```
foreverpower: Also, AtFriendLimit should check if the other player is at the max friends limit as well as the local player.

CoreScriptsRoot/CoreScripts/FriendPlayerPrompt.lua
ConorGriffin37: :IsFriendsWith uses getFriendStatus as it's first check internally. It does an additional check for the case where getFriendStatus would return Unknown.

CoreScriptsRoot/Modules/PlayerDropDown.lua
```diff
@@ -567,6 +571,56 @@ function createPlayerDropDown()
...
+				PlayerMutedEvent:Fire(player)
+			else
+				PlayerUnMutedEvent:Fire(player)
+			end
+		end
+	end)
+
+	LocalPlayer.FriendStatusChanged:connect(function(player, friendStatus)
+		if friendStatus == Enum.FriendStatus.Friend then
+			PlayerFriendedEvent:Fire(player)
```
foreverpower: Could this fire for FriendRequestSent/FriendRequestReceived as well? That would be useful for developers implementing custom playerlists.

CoreScriptsRoot/Modules/PlayerDropDown.lua
```diff
@@ -567,6 +571,56 @@ function createPlayerDropDown()
...
+		elseif friendStatus == Enum.FriendStatus.NotFriend then
+			PlayerUnFriendedEvent:Fire(player)
+		end
+	end)
+
+	StarterGui:RegisterGetCore("PlayerBlockedEvent", function() return PlayerBlockedEvent end)
```
foreverpower: These could probably just return the BindableEvent.Event properties so that developers can't try to move/fire the event objects.

CoreScriptsRoot/CoreScripts/FriendPlayerPrompt.lua
```diff
@@ -0,0 +1,192 @@
...
+
+function SendFriendRequest(playerToFriend)
+	local success = pcall(function()
+		LocalPlayer:RequestFriendship(playerToFriend)
+	end)
+	return success
+end
+
+function AtFriendLimit(player)
+	local friendCount = PlayerDropDownModule:GetFriendCountAsync(player)
```
foreverpower: This should check if friendCount = nil and return false if it is because getFriendCountAsync returns nil if the web request fails.

PR #691, opened 2017-01-24
Add throttle for receiving friend notifications.

DarraghGriffin:
I'm hoping this will mitigate some of the friend request spam issues but it's not a full solution. We need to fix this on the server.


CoreScriptsRoot/CoreScripts/NotificationScript2.lua
```diff
@@ -83,6 +83,7 @@ local EASE_DIR = Enum.EasingDirection.InOut
 local EASE_STYLE = Enum.EasingStyle.Sine
 local TWEEN_TIME = 0.35
 local DEFAULT_NOTIFICATION_DURATION = 5
+local FRIEND_REQUEST_NOTIFICATION_THROTTLE = 5 -- Time between displaying new friend requests from one unique player.
```
hockeychaos: this seems like it would be useful as a fastint

DarraghGriffin: Added the fastint

PR #701, opened 2017-01-30
MasterControl modified so that JumpRequest fires once

JHelmsAtROBLOX:
Because Humanoid.Jump was being set on each RenderStep, and because the engine changes Jump from true to false when a jump can't occur, the Humanoid.Jump value was toggling back and forth from true to false 2 or 3 times on each jump, which made the JumpRequest signal (which fires when jump becomes true) difficult for developers to use.

So checks are added here on Humanoid:GetState() to make sure a jump can occur, in addition to setting Humanoid.Jump when SetIsJumping is called by the MasterControl submodules.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
@@ -63,7 +63,10 @@ function MasterControl:Init()
 			if not humanoid then return end
 			
 			if humanoid and not humanoid.PlatformStand and isJumping then
-				humanoid.Jump = isJumping
+				local state = humanoid:GetState()
+				if state ~= Enum.HumanoidStateType.Jumping and state ~= Enum.HumanoidStateType.Freefall and state ~= Enum.HumanoidStateType.Landed then
+					humanoid.Jump = isJumping
+				end
 			end
 			
 			local adjustedMoveValue = moveValue
@@ -111,6 +114,10 @@ end
 
 function MasterControl:SetIsJumping(jumping)
 	isJumping = jumping
+	local humanoid = self:GetHumanoid()
+	if humanoid and not humanoid.PlatformStand then
+		humanoid.Jump = isJumping
+	end
 end
 
 function MasterControl:DoJump()
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/Thumbpad.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/Thumbpad.rbxmx
@@ -181,6 +181,8 @@ function Thumbpad:Create(parentFrame)
 	
 	--input connections
 	ThumbpadFrame.InputBegan:connect(function(inputObject)
+		--A touch that starts elsewhere on the screen will be sent to a frame's InputBegan event
+		--if it moves over the frame. So we check that this is actually a new touch (inputObject.UserInputState ~= Enum.UserInputState.Begin)
 		if TouchObject or inputObject.UserInputType ~= Enum.UserInputType.Touch
 			or inputObject.UserInputState ~= Enum.UserInputState.Begin then
 			return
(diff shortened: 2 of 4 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
```diff
@@ -63,7 +63,10 @@ function MasterControl:Init()
 			if not humanoid then return end
 			
 			if humanoid and not humanoid.PlatformStand and isJumping then
-				humanoid.Jump = isJumping
+				local state = humanoid:GetState()
+				if state ~= Enum.HumanoidStateType.Jumping and state ~= Enum.HumanoidStateType.Freefall and state ~= Enum.HumanoidStateType.Landed then
```
DarraghGriffin: Does this work for games that try to read the Jump value again while the humanoid is in the air to do a double jump? I think Super Bomb Survival does this.

JHelmsAtROBLOX:
The behavior should now be:

1) When an input mapped to jump on a given platform begins, Humanoid.Jump
will be set to true, and then set to false when that input ends (though the
engine will have likely have already set it back to false).

2) It will be set to true again if the jump input hasn't ended and the
humanoid re-enters a state from which jump is possible (i.e. the player
holds the spacebar to jump continuously).

So that should be a more reliable way to allow a double jump when the
humanoid is in the air, whether listening for changes to Humanoid.Jump or
UserInputService.JumpRequest.

JHelmsAtROBLOX:
Additional issue addressed:

Currently, on touch devices using Thumbstick or Thumbpad control schemes, if a touch begins in the area of the screen that allows the camera to manipulated, and then moves into the thumbstick/pad frame, the touch will be moving both the camera and the player's avatar.

With these changes to the Thumbstick and Thumbpad modules, a touch will only effect the player's movement if it begins in the thumbstick/pad frame.

DarraghGriffin:
This change makes sense, looks like InputBegan is firing because the input has begun on that particular GuiObject but it isn't actually a new input. Maybe we should change this @hockeychaos?

The other change looks good to me too, can you confirm that the steps in this wiki article will still work for Double Jumping? http://wiki.roblox.com/index.php?title=Double_Jumping

JHelmsAtROBLOX:
I can confirm that the steps in that wiki article will work for Double Jumping with these changes (however they don't work with the MasterControl script as it currently is).

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
```diff
@@ -111,6 +114,10 @@ end
 
 function MasterControl:SetIsJumping(jumping)
 	isJumping = jumping
+	local humanoid = self:GetHumanoid()
+	if humanoid and not humanoid.PlatformStand then
+		humanoid.Jump = isJumping
```
SolarCrane: Is there any reason we need to set the jump here rather than letting the renderstep function update it?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl/Thumbpad.rbxmx
```diff
@@ -181,7 +181,8 @@ function Thumbpad:Create(parentFrame)
 	
 	--input connections
 	ThumbpadFrame.InputBegan:connect(function(inputObject)
-		if TouchObject or inputObject.UserInputType ~= Enum.UserInputType.Touch then
+		if TouchObject or inputObject.UserInputType ~= Enum.UserInputType.Touch
+			or inputObject.UserInputState ~= Enum.UserInputState.Begin then
```
SolarCrane: We should probably comment on why we need to check for begin for future reference

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
JHelmsAtROBLOX: I believe so - in SetIsJumping, humanoid.Jump is being set independent of the humanoid's state, so that JumpRequest reliably fires whenever the jump button is pressed. In renderstep, it only gets set if the player is in state that allows them to jump, other wise MasterControl is fighting with the engine as to whether humanoid.Jump should be true, and JumpRequest generally fires 2 or 3 times on each jump. So you need to set it in both.

PR #721, opened 2017-02-07
Fix error in NewFollower Event 

DarraghGriffin:
An attempt to index a nil value error was reported on the devforum:
http://devforum.roblox.com/t/the-developer-console-core-script-bug-logs-they-burn/33356

Not sure where this error is coming from but adding this simple check
should resolve it.


foreverpower:
This is likely a race condition. Remotes don't send instances, only references. So if a peer sends to another peer a reference to an instance that is parented to nil or no longer exists, it would appear as nil. So more than likely, the player has left the game but because of latency, the client sending a follow request is not yet made aware of that change.

However, there shouldn't be a reason for player1 to be nil, so I assume that check can be removed. If player1 is nil for some reason, then this is a bigger issue with remoteevents that would need to be addressed.

spotco:
We should address the cause of the problem in the future. This PR will fix the specific symptoms mentioned.

PR #728, opened 2017-02-08
Harden remote events against bad input

matthewdean:
* Use typeof(value) instead of pcalling to determine userdata type
* Perform type checks on arguments supplied by clients

### DIFF
--- a/CoreScriptsRoot/CoreScripts/NotificationScript2.lua
+++ b/CoreScriptsRoot/CoreScripts/NotificationScript2.lua
@@ -450,7 +450,6 @@ local function onSendNotificationInfo(notificationInfo)
 	--Right now, calling sendNotification(...) will essentially translate into a sendNotificationInfo(table), which will then be translated back
 	--into a parameter list here. One option in the future is to update the code that produces the notification in non-VR to use notificationInfo table.
 	onSendNotification(notificationInfo.Title, notificationInfo.Text, notificationInfo.Image, notificationInfo.Duration, notificationInfo.Callback, notificationInfo.Button1Text, notificationInfo.Button2Text)
-	onSendNotification(notificationInfo.Title, notificationInfo.Text, notificationInfo.Image, notificationInfo.Duration, notificationInfo.Callback, notificationInfo.Button1Text, notificationInfo.Button2Text)
 end
 BindableEvent_SendNotificationInfo.Event:connect(onSendNotificationInfo)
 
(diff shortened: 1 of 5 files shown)


CoreScriptsRoot/CoreScripts/NotificationScript2.lua
```diff
@@ -450,6 +450,7 @@ local function onSendNotificationInfo(notificationInfo)
 	--Right now, calling sendNotification(...) will essentially translate into a sendNotificationInfo(table), which will then be translated back
 	--into a parameter list here. One option in the future is to update the code that produces the notification in non-VR to use notificationInfo table.
 	onSendNotification(notificationInfo.Title, notificationInfo.Text, notificationInfo.Image, notificationInfo.Duration, notificationInfo.Callback, notificationInfo.Button1Text, notificationInfo.Button2Text)
+	onSendNotification(notificationInfo.Title, notificationInfo.Text, notificationInfo.Image, notificationInfo.Duration, notificationInfo.Callback, notificationInfo.Button1Text, notificationInfo.Button2Text)
```
hockeychaos: is this supposed to be duplicated?

CoreScriptsRoot/CoreScripts/NotificationScript2.lua
```diff
@@ -791,7 +791,7 @@ local function createDeveloperNotification(notificationTable)
 		if type(notificationTable.Title) == "string" and type(notificationTable.Text) == "string" then
 			local iconImage = (type(notificationTable.Icon) == "string" and notificationTable.Icon or "")
 			local duration = (type(notificationTable.Duration) == "number" and notificationTable.Duration or DEFAULT_NOTIFICATION_DURATION)
-			local success, bindable = pcall(function() return (notificationTable.Callback:IsA("BindableFunction") and notificationTable.Callback or nil) end)
+			local bindable = (typeof(notificationTable.Callback) == "Instance" and notificationTable.Callback:IsA("BindableFunction") and notificationTable.Callback or nil)
```
SolarCrane: Adding `and notificationTable.Callback` seems a little superfluous since typeof() is already checking if it is nil and would not return "Instance" in that case.

DarraghGriffin: This makes the statement evaluate to notificationTable.Callback, it isn't a nill check.

(typeof(notificationTable.Callback) == "Instance" and notificationTable.Callback:IsA("BindableFunction") and notificationTable.Callback or nil)

matthewdean: In the future we should probably use if then statements instead of complicated ternary expressions

DarraghGriffin:
Can we accept this? Looks like a good change.

PR #729, opened 2017-02-08
New Chat: Fix /team showing unfiltered text to the user who had entered it.

ConorGriffin37:
This also fixes a bubble chat issue where it would not correctly display message from /w or /team for the user who had entered the message.

Basically in this case we are sending the unfiltered message for the user who had entered it but pretending it is filtered. We can't change this behaviour even if it is not ideal because it will break legacy client side partial chat copies. 

We should always still send the filtered text because the client might want to display that instead, so a user can see what parts of their message was filtered.

### DIFF
--- a/CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
+++ b/CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
@@ -65,18 +65,17 @@ function methods:SendMessageToSpeaker(message, speakerName, fromSpeaker, extraDa
 			return
 		end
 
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
```diff
@@ -65,18 +65,17 @@ function methods:SendMessageToSpeaker(message, speakerName, fromSpeaker, extraDa
 			return
 		end
 
+		-- We need to claim the message is filted even if it not in this case for compatibility with legacy client side code.
```
foreverpower: filtered is spelled wrong.

PR #730, opened 2017-02-09
Pstats.memoryv1

DougBanksPersonal:
new tab in dev console shows table of memory stats: who's using how much memory.

tried to add most new code outside of DevConsole as that is a real monstrosity. 😢 

fixed an error in DevConsole where content in scroll widget that's smaller than allotted space grows up from the bottom instead of down from the top.


CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -2258,7 +2327,11 @@ end
 ----------------
 -- Scroll bar --
 ----------------
-function Methods.ApplyScrollbarToFrame(devConsole, scrollbar, window, body, frame)
+function Methods.ApplyScrollbarToFrame(devConsole,
```
hockeychaos: I know this code has been here for awhile, but I'm curious what it does in general? Is this a Lua implementation of ScrollingFrame?

hockeychaos: If so, is there a reason we don't use ScrollingFrame? Some bug or defect or feature missing?

DougBanksPersonal: So I don't know much about what's what with GUI widgets and git.
ScrollingFrame like 
http://wiki.roblox.com/index.php?title=API:Class/ScrollingFrame?

It seems to be a lua implementation of scroll logic, yes: renders the scroll bar, the thumb, etc.

I don't know why the person who implemented this chose to roll his own vs using what's there.

DougBanksPersonal: Evidently it was some intern who is long gone.

hockeychaos: oh I didn't realize in the moment this is in dev console :(

PR #731, opened 2017-02-12
Improve PurchasePromptScript Error Message

foreverpower:
Use DEFAULT_ERROR instead of IN_GAME_PURCHASE_DISABLED if getProductInfo() fails. The current behavior is confusing because it does not distinguish between getProductInfo() failing because the marketplace is down/disabled or getProductInfo() failing because an invalid ID was specified.

Also fix missed case of game:GetService() instead of indexing the service directly, and improve warn message given when getProductInfo() fails.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
+++ b/CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
@@ -908,7 +908,7 @@ local function getProductInfo()
 	end
 
 	if not success or not result then
-		print("PurchasePromptScript: getProductInfo failed because", result)
+		warn("PurchasePromptScript: getProductInfo failed because", result, "Make sure a valid ID was specified")
 		return nil
 	end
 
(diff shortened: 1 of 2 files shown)


CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
```diff
@@ -908,7 +908,7 @@ local function getProductInfo()
 	end
 
 	if not success or not result then
-		print("PurchasePromptScript: getProductInfo failed because", result)
+		warn("PurchasePromptScript: getProductInfo failed because", result, "Make sure a valid ID was specified")
```
hockeychaos: result should contain any relevant info, this additional string does not seem to provide accurate user feedback

CoreScriptsRoot/CoreScripts/PurchasePromptScript3.lua
```diff
@@ -1015,7 +1015,7 @@ local function getProductInfo()
 	end
 
 	if not success or not result then
-		print("PurchasePromptScript: getProductInfo failed because", result)
+		warn("PurchasePromptScript: getProductInfo failed because", result, "Make sure a valid ID was specified")
```
hockeychaos: same issue here

CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
foreverpower: The only problem is that the error message provided by GetProductInfo is not clear if an invalid productId is specified. The error message is "MarketplaceService:getProductInfo() failed because rawProductInfo was empty", which does not clearly explain the issue.

PR #737, opened 2017-02-17
Chat: Fix memory leak.

ConorGriffin37:
We need to destroy the created BindableEvents to make sure that an object is not kept around because these are connected to.

We also need to disconnect the connection to the Player.Changed event, because otherwise this will leak memory.

### DIFF
--- a/CoreScriptsRoot/Modules/Server/ServerChat/DefaultChatModules/ExtraDataInitializer.lua
+++ b/CoreScriptsRoot/Modules/Server/ServerChat/DefaultChatModules/ExtraDataInitializer.lua
@@ -128,29 +128,41 @@ local function Run(ChatService)
 
 	ChatService.SpeakerAdded:connect(function(speakerName)
 		local speaker = ChatService:GetSpeaker(speakerName)
-		local player = speaker:GetPlayer()
-
-		if (not speaker:GetExtraData("NameColor")) then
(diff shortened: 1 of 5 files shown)


CoreScriptsRoot/Modules/Server/ServerChat/DefaultChatModules/ExtraDataInitializer.lua
```diff
@@ -128,29 +128,41 @@ local function Run(ChatService)
...
 			end
 		end
-		if (not speaker:GetExtraData("Tags")) then
+		if not speaker:GetExtraData("Tags") then
 			speaker:SetExtraData("Tags", {})
 		end
+	end)
 
+	-- The Player changed connections must be stored and disconnected to avoid a memory leak.
+	-- This took about a week to track down.
```
DarraghGriffin: Is this comment necessary? Also, we can leave this information here but if we fix the memory leak in the engine it will be misinforming anyone that reads the chat code, should add a note about the nature of the memory leak if adding anything at all.

ConorGriffin37: I will remove this if we fix the issue in the engine. If a comment is not left here, it will look like unnecessary optimization and may be removed.

PR #738, opened 2017-02-23
Show Visible Age

ConorGriffin37:
Show visible account age on everything but Xbox and VR. Once I get the design for those I will make a pull request with that.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/Topbar.lua
+++ b/CoreScriptsRoot/CoreScripts/Topbar.lua
@@ -70,6 +70,11 @@ while not Player do
 	Player = PlayersService.LocalPlayer
 end
 
+local accountTypeText = "Account: Over 13"
+if Player:GetUnder13() then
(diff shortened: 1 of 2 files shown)


CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1633,6 +1676,69 @@ local function CreateNotificationsIcon3D(topBarInstance, panel, menubar)
...
+		AutoButtonColor = false;
+		Image = "rbxasset://textures/ui/TopBar/Round.png";
+		ScaleType = Enum.ScaleType.Slice;
+		SliceCenter = Rect.new(10, 10, 10, 10);
+		ImageTransparency = 0;
+		BackgroundTransparency = 1;
+		Parent = container;
+	}
+
+	local accountTypeText = "Account: Over 13"
```
hockeychaos: can we do this logic in one place?

PR #740, opened 2017-02-24
Add support for chat privacy settings

ConorGriffin37:


CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -1861,8 +1867,14 @@ local function OnCoreGuiChanged(coreGuiType, coreGuiEnabled)
...
 		end
 		if showTopbarChatIcon then
-			if chatIcon then
-				AddItemInOrder(LeftMenubar, chatIcon, LEFT_ITEM_ORDER)
+			if Util.IsTouchDevice or ChatModule:IsBubbleChatOnly() then
+				if chatIcon and canChat then
+					AddItemInOrder(LeftMenubar, chatIcon, LEFT_ITEM_ORDER)
+				end
+			else
+				if chatIcon then
```
hockeychaos: we still want chat icon on non touch devices?

CoreScriptsRoot/Modules/Server/ClientChat/ChatMain.lua
```diff
@@ -1021,6 +1035,20 @@ spawn(function()
...
 end)
 
+if chatPrivacySettingsEnabled then
+	spawn(function()
+		local success, canLocalUserChat = pcall(function()
+			return Chat:CanUserChatAsync(LocalPlayer.UserId)
+		end)
+		if success then
+			canChat = canLocalUserChat
+			if canChat == false then
```
hockeychaos: is this check necessary?

CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
```diff
@@ -57,14 +57,50 @@ function methods:SendMessageObjToFilters(message, messageObj, fromSpeaker)
...
+		chatPrivacySettingsEnabled = chatPrivacySettingsValue
+	end
+	return chatPrivacySettingsEnabled
+end
+
+function methods:CanCommunicateByUserId(userId1, userId2)
+	if ChatSettingsEnabled() == false then
+		return true
+	end
+	if userId1 < 1 or userId2 < 1 then
```
hockeychaos: all guests can chat? or is this so guests can see chat?

CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
```diff
@@ -57,14 +57,50 @@ function methods:SendMessageObjToFilters(message, messageObj, fromSpeaker)
...
+	return success and canCommunicate
+end
+
+function methods:CanCommunicate(speakerObj1, speakerObj2)
+	local player1 = speakerObj1:GetPlayer()
+	local player2 = speakerObj2:GetPlayer()
+	if player1 and player2 then
+		self:CanCommunicateByUserId(player1.UserId, player2.UserId)
+	end
+	return true
```
hockeychaos: why do we always return true?

CoreScriptsRoot/CoreScripts/Topbar.lua
ConorGriffin37: Yeah, this is because there are 2 chat icons on mobile. We only want to get rid of the one that focus the chat bar, not the other one.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -70,6 +74,8 @@ while not Player do
 	Player = PlayersService.LocalPlayer
 end
 
+local canChat = true
```
DarraghGriffin: How can you guarantee that canChat has been set before the setting is read? It doesn't matter in this case because you are removing the top bar button when canChat is set but the canChat variable seems potentially misleading.

CoreScriptsRoot/Modules/Server/ClientChat/ChatMain.lua
ConorGriffin37: I think so, a developer could disable the chat bar before this is called.

CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
ConorGriffin37: This would be a problem, I will fix this.

CoreScriptsRoot/CoreScripts/Topbar.lua
```diff
@@ -2070,6 +2082,26 @@ else
 	topbarEnabledChanged()
 end
 
+if chatPrivacySettingEnabled then
+	spawn(function()
+		if Util.IsTouchDevice() or ChatModule:IsBubbleChatOnly() then
+			local success, localUserCanChat = pcall(function()
+				return ChatService:CanUserChatAsync(LocalPlayer.UserId)
```
DarraghGriffin: If the ChatService API goes down or fails this will disable chat for the user, is this the desired behavior? Could we retry if it fails?

CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
ConorGriffin37: If one player is a bot, we will return true. But this is a bug, we should also return self:CanCommunicateByUserId(player1.UserId, player2.UserId)

CoreScriptsRoot/CoreScripts/Topbar.lua
ConorGriffin37: I think this is passed in as a join param, so we don't need to worry about this

CoreScriptsRoot/CoreScripts/Topbar.lua
ConorGriffin37: We don't make that guarantee and we can't slow down the top bar setup waiting for this variable. We need to remove the top bar button when it is set.

CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
```diff
@@ -197,6 +233,22 @@ function methods:GetHistoryLog()
 	return DeepCopy(self.ChatHistory)
 end
 
+function methods:GetHistoryLogForSpeaker(speaker)
```
spotco: ?

CoreScriptsRoot/Modules/Server/ServerChat/ChatChannel.lua
```diff
@@ -197,6 +234,22 @@ function methods:GetHistoryLog()
...
+	local userId = -1
+	local player = speaker:GetPlayer()
+	if player then
+		userId = Player.UserId
+	end
+	local chatlog = {}
+	for i = 1, #self.ChatHistory do
+		local logUserId = self.ChatHistory[i].SpeakerUserId
+		if self:CanCommunicateByUserId(userId, logUserId) then
```
foreverpower: Won't this make a lot of requests, or does it automatically cache the value for all players that have joined that server? I also thought this only worked for in-game players.

PR #742, opened 2017-02-24
Show visible age xbox

ConorGriffin37:


CoreScriptsRoot/Modules/TenFootInterface.lua
```diff
@@ -339,6 +410,10 @@ local moduleApiTable = {}
 		return TenFootInterfaceModule:CreateHealthBar()
 	end
 
+	function moduleApiTable:CreateAccountType(accountTypeText)
+		return TenFootInterfaceModule:CreateAccountType(accountTypeText)
```
hockeychaos: should we check if is tenfootinterface before doing this?

ConorGriffin37: We check isTenFootInterface before this is called

PR #747, opened 2017-02-27
Replace Lua's type() with Roblox's typeof()

boynedmaster:


matthewdean:
There is no benefit to using typeof in this case. I think the general rule should be to use typeof only when necessary (e.g. to tell if a value is an Instance).

kyle-emmerich:
To expand on that, we should stick to using type() whenever we can for compatibility with non-Roblox Lua. It's confusing to beginners.

PR #749, opened 2017-02-27
Add CoreScript option "AddOption" to add custom settings.

boynedmaster:
If you modify your CoreScripts for Roblox Studio and open up this place in Edit Mode, you can test this out now. It's not copy locked.

https://www.roblox.com/games/672801155/roblox-testing-zone

```
local printEvent = Instance.new("BindableEvent")
printEvent.Event:connect(print)

game.StarterGui:SetCore("AddOption", {
	"Test Slider",
	"Slider",
	10,
	5,
	printEvent
})

game.StarterGui:SetCore("AddOption", {
	"Test Selector",
	"Selector",
	{"Foo", "Bar", "Baz"},
	2,
	printEvent
})

game.StarterGui:SetCore("AddOption", {
	"Test Dropdown",
	"DropDown",
	{"Foo", "Bar", "Baz"},
	2,
	printEvent
})
```

Demo

### DIFF
--- a/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
@@ -331,6 +331,18 @@ local function Initialize()
         return
       end
 
+      if option[3] > 10 then
+        warn("The third entry of AddOption can't be greater than 10, setting it to 10.")
+
+        option[3] = 10
+      end
+
+      if option[3] < 1 then
+        warn("The third entry of AddOption can't be less than 1, setting it to 1.")
+
+        option[3] = 1
+      end
+
       _, __, customRow = utility:AddNewRow(this,
                   option[1],
                   "Slider",
--- a/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/GameSettings.lua
@@ -61,6 +61,7 @@ local PageInstance = nil
 local LocalPlayer = Players.LocalPlayer
 local platform = UserInputService:GetPlatform()
 local overscanScreen = nil
+local customOptions = {}
 
 ----------- CLASS DECLARATION --------------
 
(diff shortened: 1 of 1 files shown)


boynedmaster:
Forgot to mention, there's no way to actually get the values, but I'll eventually add a callback function or BindableEvent as an entry for the table.

boynedmaster:
Callbacks have been added with BindableEvents.

hockeychaos:
Generally we accept PRs from the community for bug fixes, but not features unfortunately.

boynedmaster:
That's unfortunate, but also strange. Any reason why? Any chance this could be added soon?

hockeychaos:
Generally this is due to our product team at Roblox deciding in tandem with the engineering team what features we should add next (obviously community feedback is a part of this process). In order to provide a seamless experience on the platform, following a roadmap or vision of what we want is crucial to this process. Part of that roadmap I should though is listening to the community!

In terms of adding this particular feature, I am concerned it conflates what is a Roblox platform setting with what is a game specific setting. Generally it seems like the best way to logically separate them to the end user is by what we currently do; devs can just make their own UI in their games with game specific settings. I can see why you might want this feature however, as developing your own UI requires much greater effort, but we are actively working on making developing UI in Roblox much easier.

With this I should say that it is always great to see people coming up with new ideas and even going so far as to implement them, this is generally a good way to communicate to us about what new features the community would like.

boynedmaster:
From your last paragraph, do you think it is beneficial to continue making features even if it is unlikely they will be added?

hockeychaos:
I think for now, no. But we've been talking about it here now and the idea has been heard. I can't say whether or not we will go forward with implementing something like this, but its good to know there is support for this in the community

boynedmaster:
Hm, ok. If Roblox ever changes their mind on community-made features, please tell me. There are many more features I'd like to add. Also, if you're curious, I wanted my settings in the Roblox setting menus because in my game the cursor is locked in the center, so a normal GUI would not work without having to unlock the cursor.

hockeychaos:
If you have a gui on the screen that includes a button, set the "Modal" property on the button to true. This will unlock the cursor as long as that button is visible

boynedmaster:
Yes I know, but there's no good way (intuitivity wise) to implement that.

PR #751, opened 2017-02-27
Remove Lua implementation of GetTextSize and instead use TextService:GetTextSize()

ConorGriffin37:
This lua implementation is no longer necessary because TextService:GetTextSize() has been changed to userscript level. This also had a memory leak: http://devforum.roblox.com/t/bubblechat-playerscript-textsizecache-memory-leak/33608


hockeychaos:
should we flag? or does this seem ok to you?

PR #752, opened 2017-02-27
Clean up old FastFlag UseShutdownToLeave

jovannic:
Defaults to true in CPP code.

### DIFF
--- a/CoreScriptsRoot/Modules/Settings/Pages/LeaveGame.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/LeaveGame.lua
@@ -25,9 +25,6 @@ local PageInstance = nil
 RobloxGui:WaitForChild("Modules"):WaitForChild("TenFootInterface")
 local isTenFootInterface = require(RobloxGui.Modules.TenFootInterface):IsEnabled()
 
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/Settings/Pages/LeaveGame.lua
```diff
@@ -95,13 +92,9 @@ local function Initialize()
 		buttonSize = UDim2.new(0, 300, 0, 80)
 	end
 
-	this.LeaveGameButton = utility:MakeStyledButton("LeaveGame", "Leave", buttonSize, useShutdownToLeave and this.LeaveFunc or nil)
+	this.LeaveGameButton = utility:MakeStyledButton("LeaveGame", "Leave", buttonSize, this.LeaveFunc or nil)
```
hockeychaos: think you probably want to get rid of the "or nil"

jovannic: Sure I can remove that too. Doesn't hurt but it's also pointless.

jovannic:
Updated.

PR #758, opened 2017-03-01
Remove unnecessary comment. 

ConorGriffin37:
This used to be true, but a fix is being added to the engine for this.


foreverpower:
If a fix exists, the code should be removed also.

ConorGriffin37:
The code will be removed further in the future, when the fix has been turned on and it is verified to work correctly so we know it will not be turned off.

PR #763, opened 2017-03-02
Dont update DialogPanel when VR is disabled

DarraghGriffin:
We should avoid doing this update every RenderStep when VR is disabled.
I changed the bound render step function to exit early if VR is not
enabled. We could instead just not bind the function at all, Kyle can
take a look at doing that instead.


DarraghGriffin:
cc @kyle-emmerich

kyle-emmerich:
Ah, that's interesting, totally forgot about that module.

PR #767, opened 2017-03-03
Make ControlScript render step function faster

DarraghGriffin:
- Better manage the CachedHumanoid so we don't have to make sure it's in
the character when GetHumanoid is called
- Re-order if statement so humanoid.PlatformStand is not checked unless
isJumping is true.
- Cache VREnabled value so it doesn't need to be read every frame.
- Also removed some unused variables and unused code.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
@@ -68,7 +68,7 @@ local function characterAdded(character)
 	LocalCharacter = character
 	CachedHumanoid = LocalCharacter:FindFirstChildOfClass("Humanoid")
 	characterAncestryChangedConn = character.AncestryChanged:connect(function()
-		if not character:IsDescendantOf(workspace) then
+		if character.Parent == nil then
 			LocalCharacter = nil
 		else
 			LocalCharacter = character
@@ -92,6 +92,7 @@ function MasterControl:Init()
 
 	local getHumanoid = self.GetHumanoid
 	local moveFunc = LocalPlayer.Move
+	local getUserCFrame = VRService.GetUserCFrame
 	local renderStepFunc = function()
 		local humanoid = getHumanoid()
 		if not humanoid then return end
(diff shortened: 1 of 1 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
```diff
@@ -12,95 +12,115 @@
...
+	end
+
+	if characterChildRemovedConn then
+		characterChildRemovedConn:disconnect()
+	end	
+	
+	LocalCharacter = character
+	CachedHumanoid = LocalCharacter:FindFirstChildOfClass("Humanoid")
+	characterAncestryChangedConn = character.AncestryChanged:connect(function()
+		if not character:IsDescendantOf(workspace) then
```
hockeychaos: when does this happen? do we need to worry about this case? should this really be, if the character has no parent, lets stop tracking it?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
```diff
@@ -12,95 +12,115 @@
...
 			end
-			
-			moveFunc(LocalPlayer, adjustedMoveValue, true)	
-			
-			
 		end
+		
+		local adjustedMoveValue = moveValue
+		if VREnabled and workspace.CurrentCamera.HeadLocked then
+			local vrFrame = VRService:GetUserCFrame()
```
hockeychaos: should we cache this function lookup?

hockeychaos: also is this equivalent to UserInputService.UserHeadCFrame?

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
DarraghGriffin: It only happens if the character is re-parented to be a descendant of another top level container by the developer. This case probably doesn't matter, we can just check character.Parent == nil here.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/ControlScript/MasterControl.rbxmx
DarraghGriffin: UserHeadCFrame is deprecated in favor of this method now, I'll cache the lookup though

PR #768, opened 2017-03-03
Prevent Panel3D updating when VR is not enabled

DarraghGriffin:
We can unbind and re-bind the render step function when the VREnabled
property changes to make sure this only runs when it needs to.


DarraghGriffin:
cc @kyle-emmerich

PR #772, opened 2017-03-06
Legacy partial copy versions of the chat are broken by the new update

ConorGriffin37:
Partial copies versions of the chat that copy ChatScript but do not copy the MessageCreatorModules are broken by the new update to the chat system. RegisterGuiRoot is removed from MessageCreatorUtil and old versions of the chat still expect this to exist. This bug causes chat to be broken in meep city.

CLI-15602


CoreScriptsRoot/Modules/Server/ClientChat/DefaultClientChatModules/MessageCreatorModules/Util.lua
```diff
@@ -304,6 +304,10 @@ function methods:NewBindableEvent(name)
 	return bindable
 end
 
+function methods:RegisterGuiRoot()
+	-- This is left here for compatibility with legacy chat versions.
```
spotco: Add a warn

PR #778, opened 2017-03-10
Update backpack to use UIGridLayout

ConorGriffin37:


kyle-emmerich:
Quite the change! :+1:

DarraghGriffin:
How is the new code longer than the old code? Shouldn't this simplify it??

PR #780, opened 2017-03-13
Have camera type modules return singletons

Fraktality:
Currently, external scripts have no way of interfacing with the camera type modules because those modules return factories for their respective camera types.  This PR aims to fix that by having those modules return singletons instead.

This allows developers to elegantly interface with the camera scripts, e.g.

`require(ClassicCamera):ZoomCamera(10)`


kyle-emmerich:
Hm, I'm sort of on the fence on accepting this. We're planning some changes to the camera scripts to support some new features soon, and we'd probably prefer to break some API since we currently control the only uses of it.

In general, I think this is a good move, but it's not quite the time for it yet.

I'd also like to change these a bit more so that they are tied to a specific camera object and have more predictable behavior when switching cameras and switching input types.

hockeychaos:
I agree with @kyle-emmerich. Policy of the repo is to review + accept bug fixes, but not new functionality, unless from an internal employee. This is a good idea I think, or at the very least we want to move in this direction.

Fraktality:
That's understandable. Any API-level changes to RootCamera should happen before these are exposed.
Policy noted.

PR #800, opened 2017-03-29
CLIAVATAR-85: Make Camera follow Head part when the player is dead

DarraghGriffin:
Seems like this is easier for users to understand than the Camera
following the invisible HumanoidRootPart

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
@@ -266,15 +266,15 @@ local function CreateCamera()
 	
 	-- HumanoidRootPart when alive, Head part when dead
 	local function getHumanoidPartToFollow(humanoid, humanoidStateType)
-		if humanoidStateType ~= STATE_DEAD then
-			return humanoid.Torso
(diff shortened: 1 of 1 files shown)


hockeychaos:
this normally follows the head now right? Why are we changing this?

DarraghGriffin:
It actually follows the HumanoidRootPart with some offset to account for how much higher the head is. This can be really weird when dead when the camera is following an invisible part.

This is a good point though, I should just change this PR to make the camera follow the Head when dead because it's the same for both R6 and R15.

PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/RootCamera.rbxmx
```diff
@@ -260,6 +261,22 @@ local function CreateCamera()
 	function this:GetRenderCFrame(part)
 		GetRenderCFrame(part)
 	end
+	
+	local STATE_DEAD = Enum.HumanoidStateType.Dead
+	
+	-- HumanoidRootPart when alive, Head part when dead
+	local function getHumanoidPartToFollow(humanoid, humanoidStateType)
+		if humanoidStateType ~= STATE_DEAD then
```
hockeychaos: can we flip this? doing a negative comparison generally makes stuff harder to read, especially when it seems easy to just use ==

PR #803, opened 2017-03-30
CLIXBOX-1154: Playerlist background image fails to load on Xbox

ClaireInRoblox:
CLIXBOX-1154: Playerlist background image fails to load on Xbox. Also fixed background image is in front of the player icon.
CLIXBOX-681：Enable D-Pad navigation for Radial Menu


ClaireInRoblox:
CLIXBOX-681：Enable D-Pad navigation for Radial Menu

PR #806, opened 2017-04-01
CLIXBOX-681: Enable D-Pad navigation for Radial Menu.

ClaireInRoblox:
Enabled multi-touch buttons

### DIFF
--- a/CoreScriptsRoot/CoreScripts/GamepadMenu.lua
+++ b/CoreScriptsRoot/CoreScripts/GamepadMenu.lua
@@ -568,14 +568,14 @@ local function setupGamepadControls()
 			
 			if input.KeyCode == Enum.KeyCode.DPadUp or input.KeyCode == Enum.KeyCode.DPadDown then
 				if D_PAD_BUTTONS[Enum.KeyCode.DPadLeft] then
-					--left up
-					inputVector = Vector2.new(-1, 1).unit
+					--left up(button down) or nothing(button up)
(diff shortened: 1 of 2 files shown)


CoreScriptsRoot/CoreScripts/GamepadMenu.lua
```diff
@@ -31,6 +31,14 @@ local isTenFootInterface = tenFootInterface:IsEnabled()
 local radialButtons = {}
 local lastInputChangedCon = nil
 
+local D_PAD_BUTTONS =
```
SolarCrane: Keeping tack of input state may lead to some strange bugs, while less efficient, perhaps using http://wiki.roblox.com/index.php?title=API:Class/UserInputService/GetGamepadState may be a more solid solution?

SolarCrane:
@hockeychaos What do you think about the big picture design here? I feel like our input system should be able to supply an input vector or give the necessary functionality to create a radial menu like this without having to write a bunch of lua code.

hockeychaos:
@SolarCrane what does that mean though? Like you should be able to query the input vector from left thumbtack and dpad? If so, what do we do if someone pressing up on the thumbstick and down on the dpad? or left on the pad?

Radial Menu is tough to implement right now. We are going to expose Rotation Point soon though which may make this a lot easier.

SolarCrane:
Navigating from one icon to the next icon is provided by default in the engine. Could we foresee adding something to help people navigate from a fixed point such as in the radial menu? Obviously we don't want it to be usable only for a specific case like this.

PR #821, opened 2017-04-11
CLIAVATAR-33: Fix shift lock allowing players to see through walls

DarraghGriffin:
This change goes along with a change to the implementation of
Camera:GetLargestCutoffDistance() which returns a maximum cutoff
distance if the camera focus is inside another part. If the focus moves
in another part this change pops the camera in fully and moves the focus
to the subject position rather than the shift lock offset position.

### DIFF
--- a/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/PopperCam.rbxmx
+++ b/PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/PopperCam.rbxmx
@@ -15,6 +15,7 @@ local PopperCam = {} -- Guarantees your players won't see outside the bounds of 
 -----------------
 
 local POP_RESTORE_RATE = 0.3
+local MIN_CAMERA_ZOOM = 0.5
 
 local VALID_SUBJECTS = {
 	'Humanoid',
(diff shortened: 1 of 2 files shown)


PlayerScripts/PlayerScriptsPlace/StarterPlayer/StarterPlayerScripts/CameraScript/PopperCam.rbxmx
```diff
@@ -167,6 +170,17 @@ function PopperCam:Update()
...
 
 		LastZoomLevel = zoomLevel
+		
+		-- Stop shift lock being able to see through walls by manipulating Camera focus inside the wall
+		if EnabledCamera and EnabledCamera:GetShiftLock() and not EnabledCamera:IsInFirstPerson() then
+			if EnabledCamera:GetCameraActualZoom() < 1 then
+				local subjectPosition = EnabledCamera.lastSubjectPosition 
+				if subjectPosition then
+					Camera.Focus = CFrame_new(subjectPosition)
+					Camera.CFrame = CFrame_new(subjectPosition - MIN_CAMERA_ZOOM*EnabledCamera:GetCameraLook(), subjectPosition)
```
hockeychaos: is it possible to get both position and lookAt to be the same value? I think this results in a white screen if this happens.

DarraghGriffin: GetCameraLook will always be a unit Vector3 so the position and the lookAt in the CFrame constructor can never be the same.

PR #825, opened 2017-04-13
Fix portrait mode error

ConorGriffin37:


CoreScriptsRoot/Modules/Settings/Pages/Players.lua
```diff
@@ -474,7 +474,7 @@ local function Initialize()
 		end)
 
 		local extraOffset = 20
-		if utility:IsSmallTouchScreen() or utility:IsPortrait() then
+		if utility:IsSmallTouchScreen() or (enablePortraitMode and utility:IsPortrait()) then
```
hockeychaos: wouldn't it be better to have this check inside utility:IsPortrait?

PR #839, opened 2017-04-21
Remove JSON parser from RbxUtility

Fraktality:
Replace the 500-line JSON parser with direct calls to the HttpService methods.


DarraghGriffin:
Overall I think this is a good change to make because it removes a library we don't need but there is no guarantee that the HttpService methods have the exact same behavior as the current JSON parser in RbxUtility.

I don't think we should remove this without a flag because it could break games in unexpected ways.

Fraktality:
Turns out that there's a functional difference between the RbxUtility and HttpService parsers. RbxUtility handles embedded zeros correctly (encodes as \u0000), but HttpService treats them as terminators.

There are probably people who rely on RbxUtility not being buggy, so I'll close this for now until the HttpService issues are resolved.

Fraktality:
The HttpService issues have been fixed.

DarraghGriffin:
I think we should make this change, but not without having a flag to disable the change if there are unintended effects.

Fraktality:
Sounds good, let me know when you have the flag's name so I can add it to the PR.

DarraghGriffin:
Alright, I will add the flag "RbxUtilityRemoveJsonLibrary" if you add this to the pull request.

Fraktality:
While we're at it, do you think there's still any reason to keep the UserRbxUtilityCreateSetParentLast flag around?

kyle-emmerich:
Nah, I think we can remove that old flag safely.

CoreScriptsRoot/Libraries/RbxUtility.lua
```diff
@@ -11,7 +15,500 @@ local rbxUtilitySetParentLast = (rbxUtilitySetParentLastFlagSuccess == true and
 ------------------------------------------------------------------------------------------------------------------------
 ------------------------------------------------------------------------------------------------------------------------
 
-do
+
+if rbxUtilityRemoveJsonLibrary then
```
DarraghGriffin: Should be "if not rbxUtilityRemoveJsonLibrary then"

hockeychaos:
will remove this soon, need to resolve some merges

PR #845, opened 2017-04-26
Fix rootCamera assumption that it knows about all inputs

ConorGriffin37:


hockeychaos:
I think we need to add this to content/scripts/PlayerScripts/StarterPlayerScripts as well

PR #848, opened 2017-04-27
Fix bug where bottom bar bindings are done when the settings menu isn't visible

kyle-emmerich:


CoreScriptsRoot/Modules/Settings/SettingsHub.lua
```diff
@@ -559,7 +559,7 @@ local function CreateSettingsHub()
 			--to update, or we can just avoid using it so soon.
 			RunService.Heartbeat:wait()
 
-			if shouldShowBottomBar() then
+			if this.Visible and shouldShowBottomBar() then
```
hockeychaos: should this check be in shouldShowBottomBar?

kyle-emmerich: I thought about it but I don't think it makes sense. You probably shouldn't even be asking if you can show the bottom bar if the menu isn't already visible.

hockeychaos: so how are we getting to this code then?

kyle-emmerich: This callback always runs. :( I was going to submit another PR shortly after this that refactors it so that it runs only while the menu is visible and runs it once when the menu becomes visible.

kyle-emmerich: But I could just do that now and put it all in one PR if you'd like.

hockeychaos: yeah lets do the one

PR #855, opened 2017-05-02
Clean up FastFlag UseSetTextFromInputForVrKeyboard

jovannic:


CoreScriptsRoot/Modules/VR/VirtualKeyboard.lua
```diff
@@ -971,15 +968,7 @@ local function ConstructKeyboardUI(keyboardLayoutDefinitions)
 	end
 
 	local function UpdateTextEntryFieldText(newText)
-		local textSet = false
-		if useSetTextFromInput then
-			textSet = pcall(function() textEntryField:SetTextFromInput(newText) end)
-		end
-
-		if not textSet then
```
hockeychaos: do we no longer need textSet to do this?

jovannic: Yeah, textset was just for seeing if the flag was set and SetTextFromInput succeeded. If not, do the old thing.

PR #856, opened 2017-05-02
Pstats.analytics

DougBanksPersonal:
Using new "trackEvent" to send GoogleAnalytics events tracking usage of DeveloperConsole.


CoreScriptsRoot/Modules/DeveloperConsoleModule.lua
```diff
@@ -2219,6 +2229,22 @@ end
...
 ----------
+function Methods.GetCurrentOpenTab(devConsole)
+	local tabs = devConsole.Tabs
+	if tabs == nil then 
+		return nil
+	end
+
+	for i = 1, #tabs do
+		local tab = tabs[i]
+		if tab.Open then
```
hockeychaos: it is interesting that it could be possible to have more than one open tab... at least in theory. Would it make more sense for open tab to be a singleton value instead of a parameter on tab?

DougBanksPersonal:
> it is interesting that it could be possible to have more than one open
> tab... at least in theory. Would it make more sense for open tab to be a
> singleton value instead of a parameter on tab?
>
Yes, but/and that would be a separate CL.
Overall I am not fan of how this file was written.  1000s of lines of
undocumented crazy-person code.

Given that the punchline of such an error would be slightly inaccurate
statistics, I am OK with leaving it as-is for now.

hockeychaos:
ah yes, I forgot this was DeveloperConsole for a second there.....

PR #857, opened 2017-05-03
Only fade in/out dynamic thumbstick background once per orientation

hockeychaos:
Dynamic thumbtack should only show thumbstick area once per orientation, as it gets annoying otherwise.


kyle-emmerich:
Dynamic thumbtack?

PR #863, opened 2017-05-10
CLIAVATAR-167: Stop player thumbnails from being NA 

DarraghGriffin:
This effects the avatar thumbnails in the social notifications and the
menu player list thumbnails.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/FriendPlayerPrompt.lua
+++ b/CoreScriptsRoot/CoreScripts/FriendPlayerPrompt.lua
@@ -23,6 +23,23 @@ local PlayerDropDownModule = require(CoreGuiModules:WaitForChild("PlayerDropDown
 local THUMBNAIL_URL = "https://www.roblox.com/Thumbs/Avatar.ashx?x=200&y=200&format=png&userId="
 local BUST_THUMBNAIL_URL = "https://www.roblox.com/bust-thumbnail/image?width=420&height=420&format=png&userId="
 
+local REGULAR_THUMBNAIL_IMAGE_SIZE = Enum.ThumbnailSize.Size420x420
+local CONSOLE_THUMBNAIL_IMAGE_SIZE = Enum.ThumbnailSize.Size180x180
+
+local REGULAR_THUMBNAIL_IMAGE_TYPE = Enum.ThumnailType.AvatarBust
+local CONSOLE_THUMBNAIL_IMAGE_TYPE = Enum.ThumbnailType.AvatarThumbnail
+
+function createFetchImageFunction(...)
+	return function(imageLabel)
+		spawn(function()
+			local imageUrl = SocialUtil.GetPlayerImage(...)
+			if imageLabel and imageLabel.Parent then
+				frame.Icon.Image = imageUrl
+			end
+		end)
+	end
+end
+
 function SendFriendRequest(playerToFriend)
 	local success = pcall(function()
 		LocalPlayer:RequestFriendship(playerToFriend)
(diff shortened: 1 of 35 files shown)


ConorGriffin37:
Should the Block/FriendPlayerPrompts also be updated to use the new API?

CoreScriptsRoot/Modules/SocialUtil.lua
```diff
@@ -0,0 +1,124 @@
...
+	[Enum.ThumbnailType.AvatarThumbnail] = "https://www.roblox.com/avatar-thumbnail/image?width=%d&height=%d&format=png&userId=%d"
+}
+
+local GET_PLAYER_IMAGE_DEFAULT_TIMEOUT = 5
+local DEFAULT_THUMBNAIL_SIZE = Enum.ThumbnailSize.Size48x48
+local DEFAULT_THUMBNAIL_TYPE = Enum.ThumbnailType.AvatarThumbnail
+local GET_USER_THUMBNAIL_ASYNC_RETRY_TIME = 1
+
+--[[ Utility ]]--
+local function CreateSignal()
```
ConorGriffin37: This function is already included in multiple places in the code base. Can we end the code duplication and consolidate these to one place?

CoreScriptsRoot/Modules/SocialUtil.lua
```diff
@@ -0,0 +1,124 @@
...
+	return sig
+end
+
+--[[ Functions ]]--
+
+-- The thumbanil isn't guaranteed to be generated, this will just create the url using string.format and immediately return it.
+function SocialUtil.GetFallbackPlayerImageUrl(userId, thumbnailSize, thumbnailType)
+	local sizeNumber = THUMBNAIL_SIZE_MAP[thumbnailSize]
+	if not sizeNumber then
+		warn("SocialUtil.GetPlayerImage: No thumbnail size in map for " ..tostring(thumbnailSize))
```
ConorGriffin37: This function will warn if thumbnailSize is nil instead of just using the default. This is inconsistent with SocialUtil.GetPlayerImage.

CoreScriptsRoot/Modules/SocialUtil.lua
```diff
@@ -0,0 +1,124 @@
...
+	end
+
+	return sig
+end
+
+--[[ Functions ]]--
+
+-- The thumbanil isn't guaranteed to be generated, this will just create the url using string.format and immediately return it.
+function SocialUtil.GetFallbackPlayerImageUrl(userId, thumbnailSize, thumbnailType)
+	local sizeNumber = THUMBNAIL_SIZE_MAP[thumbnailSize]
```
ConorGriffin37: This function will error if thumbnailSize is nil. This is inconsistent with GetPlayerImage which uses the default thumbnail size in the case that it is nil. Should this function do that instead?

DarraghGriffin: This won't error and it will use the default value. It will print a warning though.

CoreScriptsRoot/Modules/SocialUtil.lua
DarraghGriffin: I could use the the definition from the Utility module in Settings but that seemed weird to me, that Module is very Settings specific and doesn't look like it should have external dependencies.

DarraghGriffin:
@ConorGriffin37 
Yes, the Block/FriendPlayerPrompts could also be update to use this, thanks for volunteering :+1:

CoreScriptsRoot/Modules/SocialUtil.lua
ConorGriffin37: Maybe we should have more general util file. This will the 6th place this code is duplicated in the code base though.
1) Chat.lua
2) ChatUtil.lua
3) PlayerDropDown.lua
4) ChatMain.lua
5) Settings/Utility.lua

CoreScriptsRoot/CoreScripts/FriendPlayerPrompt.lua
```diff
@@ -145,6 +170,8 @@ function DoPromptUnfriendPlayer(playerToUnfriend)
 					CancelActive = false,
 					Image = BUST_THUMBNAIL_URL ..playerToUnfriend.UserId,
```
ConorGriffin37: These images could be updated to get the fallback image from SocialUtil

PR #867, opened 2017-05-11
Use GetPropertyChangedSignal for Camera.Changed listeners

DarraghGriffin:
This reduces the amount of overhead when setting busy properties on
Camera such as CFrame in the CameraScripts.

### DIFF
--- a/CoreScriptsRoot/Modules/VR/Panel3D.lua
+++ b/CoreScriptsRoot/Modules/VR/Panel3D.lua
@@ -1052,24 +1052,22 @@ RunService.Heartbeat:connect(onHeartbeat)
 
 
 local cameraChangedConnection = nil
-local function onCameraChanged(prop)
-	if prop == "HeadScale" then
-		pcall(function()
-			currentHeadScale = workspace.CurrentCamera.HeadScale
-		end)
-		for i, v in pairs(panels) do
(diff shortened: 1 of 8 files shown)


kyle-emmerich:
Looks good, but we should wait until the flag is going to be removed to merge it. If we got it into the same CI commit or later, perfect. Otherwise GT sites will break. :(

CoreScriptsRoot/Modules/VR/Panel3D.lua
```diff
@@ -1052,24 +1052,22 @@ RunService.Heartbeat:connect(onHeartbeat)
...
-local function onCameraChanged(prop)
-	if prop == "HeadScale" then
-		pcall(function()
-			currentHeadScale = workspace.CurrentCamera.HeadScale
-		end)
-		for i, v in pairs(panels) do
-			v:OnHeadScaleChanged(currentHeadScale)
-		end
+local function onHeadScalePropChanged(prop)
+	pcall(function()
```
kyle-emmerich: No big deal, but we can probably remove this pcall now.

hockeychaos:
we good to merge?

DarraghGriffin:
Going to commit this with the new system.

PR #872, opened 2017-05-16
CLIPLAYEREX-99 It is possible to select players without function

garnold24:
This change stops the highlighting of player names in the escape menu,
which led users to think clicking on them would open something.

### DIFF
--- a/CoreScriptsRoot/Modules/Settings/Pages/Players.lua
+++ b/CoreScriptsRoot/Modules/Settings/Pages/Players.lua
@@ -439,7 +439,8 @@ local function Initialize()
 		nameLabel.Size = UDim2.new(0, 0, 0, 0)
 		nameLabel.ZIndex = 3
 		nameLabel.Parent = frame
-
+		
+		
 		return frame
 	end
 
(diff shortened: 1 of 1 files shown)


CoreScriptsRoot/Modules/Settings/Pages/Players.lua
```diff
@@ -440,13 +440,14 @@ local function Initialize()
...
-			frame.ImageTransparency = FRAME_DEFAULT_TRANSPARENCY
-		end)
 
+		--[[frame.MouseEnter:connect(function()
+				frame.ImageTransparency = FRAME_SELECTED_TRANSPARENCY
+			end)
+			frame.MouseLeave:connect(function()
+				frame.ImageTransparency = FRAME_DEFAULT_TRANSPARENCY
+			end)
+		end]]
```
ConorGriffin37: We leave this code here commented out?

garnold24: I have bad experiences with removing code and readding it. ;) I suppose it could just be removed since it isn't too important.

PR #874, opened 2017-05-17
CLIPLAYEREX-458 Dynamic Thumbstick graphic displays past control zone

garnold24:
The code removed was responsible for making the dynamic thumbstick not
display past the bottom half of the screen. This change will make it so
the graphic will now follow the touch input around no matter where it
is on the screen.


kyle-emmerich:
sry dan

My changes in a new PR make this one unnecessary. We can close this one.

garnold24:
;(

PR #885, opened 2017-05-28
Prevent redundant SetCore CoreGuiChatConnection attempt

Jody7:
SetCore fails when in VR Mode. Another (fix) could be preventing DoEverything being called when in VR mode, but I am unsure if it is used in VR.

### DIFF
--- a/CoreScriptsRoot/Modules/Server/ClientChat/ChatScript.lua
+++ b/CoreScriptsRoot/Modules/Server/ClientChat/ChatScript.lua
@@ -102,19 +102,17 @@ local function DoEverything()
 end
 
 function SetCoreGuiChatConnections(containerTable)
-	if not UserInputService.VREnabled then
-		local tries = 0
-		while tries < MAX_COREGUI_CONNECTION_ATTEMPTS do
-			tries = tries + 1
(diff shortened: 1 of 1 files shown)


DarraghGriffin:
@ConorGriffin37 can you take a look at this?

ConorGriffin37:
We should probably move the VR check to line 130 where the GuiService:IsTenFootInterface() check is, as we do not want to do anything with this chat in VR.

Jody7:
VR Check line changed

CoreScriptsRoot/Modules/Server/ClientChat/ChatScript.lua
```diff
@@ -125,7 +125,7 @@ function checkBothChatTypesDisabled()
 	return false
 end
 
-if not GuiService:IsTenFootInterface() then
+if (not GuiService:IsTenFootInterface()) or (not game:GetService('UserInputService').VREnabled) then
```
ConorGriffin37: This should be and, not or. Because we only want to do this in the case where it is not Vr and not ten foot interface.

Jody7:
Fixed

Issue #998, opened 2017-07-21
Unicode Charachaters no longer render in the chat bubble, but render correctly in the chat history.


stillunt1tled:
## Summary
Non-English UTF-8 Compatible characters do not render correctly (render as �) in the chat bubble that appears above a player's head.

**EDIT: This also happens with the text in /me. For example: '/me likes the letter Д' will produce this:** ![screenshot of result](https://i.gyazo.com/14ea7ba18bf62e399a2b590a28082de4.png)

## Reproduction
To reproduce this bug in game, open a game in the PLAYER that has 'Both' as the chat mode. For this example we will chat the character U+0414 (Cyrillic Capital Letter De, Д) and it will produce the following: 

**NOTE: While the chat history faded out before I could take the screenshot, it did render correctly in the chat history!**

![robloxscreenshot20170721_084924401](https://user-images.githubusercontent.com/28220558/28464078-aa49b590-6df1-11e7-83ed-d1f22694d301.png)

## What about emojis?
They're fine! Don't worry about your pet snake!

![robloxscreenshot20170721_085132459](https://user-images.githubusercontent.com/28220558/28464144-e3ace53c-6df1-11e7-95b1-aca5fb433277.png)

## Possible Fixes
I do believe the problem is caused by the use of font. Some fonts will have correct characters for some special characters, but some of them don't and replace them with said box with a X through it (also known as �, just different fonts display it differently)

### Thank you for reading, please discuss this below! (would appreciate if someone turned this into a pull request!)

*edit: oh noes we're so close to issue #1000*

stillunt1tled:
I just did some digging, and it is the font that is the problem. I changed the Me command to use SourceSansBold, and it correctly displayed unicode characters. I will do some more coding to find a fix and make a pull request when I am done and it is all working.

stillunt1tled:
This issue is now **Pull Request** #999

PR #999, opened 2017-07-21
Change BubbleChat font from SourceSans to SourceSansBold

stillunt1tled:
Solves issue #998 

SourceSans cannot correctly display Non-English Unicode Characters, and instead replaces them with � (SourceSans displays � as a box with a X through it). SourceSansBold, however, can display Non-English Unicode Characters. Arial can also do so, so if SourceSansBold is considered ugly or out-of-place, Arial can be used to replace it.

I suggest reverting this change (or keeping it, because SourceSansBold looks pretty nice) once ROBLOX fixes the font to work with unicode.

Want to see how it would look? Vist [my testing ground](https://www.roblox.com/games/929781001/github-issue-998-pull-link) to view how this would look.

*p.s*
picture of sourcesansbold being used in-game

### DIFF
--- a/CoreScriptsRoot/Modules/Server/ClientChat/BubbleChat/BubbleChat.lua
+++ b/CoreScriptsRoot/Modules/Server/ClientChat/BubbleChat/BubbleChat.lua
@@ -20,7 +20,7 @@ end
 local PlayerGui = LocalPlayer:WaitForChild("PlayerGui")
 
 --[[ SCRIPT VARIABLES ]]
-local CHAT_BUBBLE_FONT = Enum.Font.SourceSans
+local CHAT_BUBBLE_FONT = Enum.Font.SourceSansBold -- changed because SourceSans cannot display certain unicode charachters correctly
 local CHAT_BUBBLE_FONT_SIZE = Enum.FontSize.Size24 -- if you change CHAT_BUBBLE_FONT_SIZE_INT please change this to match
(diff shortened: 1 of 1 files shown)


stillunt1tled:
I'm still a bit torn on Arial or SourceSansBold.

PhoenixShay:
Personally, I think it would be better to use Arial instead of a bold font if ROBLOX even wants people using the characters that aren't supported.

stillunt1tled:
RE: Your comment, @PhoenixShay, Arial has sizing issues. It was either too big or too small (Size24 was too big and looked like it was trying to escape the Chat bubble, Size18 was too small and hard to read, for me at least). SourceSansBold has no sizing issues because it is still SourceSans. I would love to hear if you can fix this problem!

stillunt1tled:
Also, Its not that ROBLOX doesn't want people to use unsupported characters, It's that supported characters are not correctly rendered.

VRService:
Frankly this is a user preference for than a requirement.

stillunt1tled:
Maybe I (or someone else) could make a pull where new settings added to the escape menu involving font.

Jody7:
Major changes to UX should be left to internal ROBLOX Staff IMHO

hockeychaos:
generally this is something we have our design team internally iterate on

stillunt1tled:
understandable, but now I have to deal with being unable to speak to my foreign friend in bubble-chat only games because reasons

Issue #1004, opened 2017-08-01
Backpack Menu stays open when tool is removed while dragging a tool


Jody7:
If you are moving a tool in your backpack and the tool is removed, the backpack will stay open as it still thinks the tool is being moved.

Jody7:
https://gyazo.com/095704424450b5f3812e92826316a3a8

cliffchapmanrbx:
Hi there, thank you for your feedback. We're going to be closing this repository as we've chosed [the Roblox Dev Forum](https://devforum.roblox.com/) as our venue for developer feedback. Please feel free to migrate your feedback to a DevForum post if you feel it is still relevant.

PR #1005, opened 2017-08-01
Handle Tool Parent change while Dragging in Backpack

Jody7:
From this: https://github.com/Roblox/Core-Scripts/issues/1004

When the Parent of a Tool is changed, and you are still dragging it in backpack, it is not properly handled as DraggingStopped


hockeychaos:
can't merge with the conflicts, I think this issue may of been fixed

Jody7:
problem still exists, will resolve conflicts

Issue #878, opened 2017-05-19
PromptProductPurchaseFinished is depreciated, but PromptPurchaseFinished does not fire for Developer products


Quenty:
The event `PromptProductPurchaseFinished` is depreciated, but `PromptPurchaseFinished` does not fire for Developer products. This means that `PromptProductPurchaseFinished` must be used to detect products finishing purchase, forcing depreciated code use. 

The code causing it:
https://github.com/Roblox/Core-Scripts/blob/0376349c8547c027de58a4e1dd1ffd12051292cf/CoreScriptsRoot/CoreScripts/PurchasePromptScript3.lua#L945

* http://wiki.roblox.com/index.php?title=API:Class/MarketplaceService/PromptProductPurchaseFinished
* http://wiki.roblox.com/index.php?title=API:Class/MarketplaceService/PromptPurchaseFinished

blujagu:
MarketplaceService.ProcessReceipt is the replacement for detecting purchase completion, so your issue should be covered. However, this is a major issue for a properly constructed UI that wants to disable it's "Buy" buttons until the server returns something. **If a player presses "Cancel", the ONLY event to listen for is PromptProductPurchaseFinished, so it being deprecated is not a good thing until there is a fitting replacement for capturing the cancel button event.**

Quenty:
Yes. To clarify:`MarketplaceService.ProcessReceipt` only is called when a purchase is _made_ (i.e. the user has bought the product). I need to know when a purchase is not made too.

The solution is to make PromptPurchaseFinished fire on DeveloperProducts when cancelled.

ArceusInator:
![dlzapoo](https://user-images.githubusercontent.com/2127060/28246469-7c04d730-69cf-11e7-93e9-c59856529191.png)
![dlzapoo](https://user-images.githubusercontent.com/2127060/28246469-7c04d730-69cf-11e7-93e9-c59856529191.png)
![dlzapoo](https://user-images.githubusercontent.com/2127060/28246469-7c04d730-69cf-11e7-93e9-c59856529191.png)
![dlzapoo](https://user-images.githubusercontent.com/2127060/28246469-7c04d730-69cf-11e7-93e9-c59856529191.png)
![dlzapoo](https://user-images.githubusercontent.com/2127060/28246469-7c04d730-69cf-11e7-93e9-c59856529191.png)
[...]

PR #1006, opened 2017-08-03
Address #878. Fire SignalPromptPurchaseFinished on developer products

Quenty:
Address issue #878: PromptProductPurchaseFinished is depreciated, but PromptPurchaseFinished does not fire for Developer products

#878


ghost (changes requested):
~~I don't see how this solves #878. It will actually break non-developer product purchases since PromptPurchaseFinished will not fire if IsPurchasingConsumable.~~ I'm stupid. I didnt see the line of code under the If statement. But I still dont see how it solves #878.

Quenty:
It lets `PromptPurchaseFinished` fire for both normal and developer products, providing an official API. 

It's just firing an event when `PromptPurchaseFinished` occurs.

There should be another event to fire when a prompt shows up, but that's technically a new feature.

stillunt1tled:
@Quenty there are conflicts, please resolve.

cliffchapmanrbx:

PR #1007, opened 2017-08-06
RobloxGui/Modules variable usage

Wittiest:
A few times CoreGui.RobloxGui was used when RobloxGui was already initialized.

TopbarConstants had an incorrect comment title.


stillunt1tled:
This PR has conflicts, please resolve them.

Also, some CoreScripts use `CoreGui.RobloxGui.Modules` or `RobloxGui.Modules` when `Modules` is already defined and initalized. I do think you should fix this too.

Thanks for helping out :+1:

Wittiest:
Done.

hockeychaos:
Looks like there are conflicts, will close for now but can look at if these issues still persist

Wittiest:
I could have fixed the conflicts. Though, I'm sure this sort of PR isn't vital to your code so unless requested I will not resubmit it

PR #1011, opened 2017-08-07
Lua Chat Client Custom Tag Handling

Xsitsu:
Client side of Lua Chat can now handle tags being added to speaker messages. They are displayed to the left side of a speaker's name.


hockeychaos:
merged internally

PR #1014, opened 2017-08-12
Fix character not being ignored in first person with invisicam enabled

Quenty:
Character is not ignored in first person with invisicam enabled causing an issue where in first person you can see your character's body as invisicam glitches out with the the TransparencyController not updating every frame.

This only happens when running or when the character is moved a bit from the camera origin. 

This is clearly a bug. This one change can fix that.


Quenty:
See issue #1015

hockeychaos:
we've merged a fix into the internal version of this that essentially does the same thing, albeit with a range check to make sure you are reasonably close to the player before we start ignorning it. will show up here when we merge back in

Quenty:
Thanks!

Issue #1016, opened 2017-08-13
Custom Icon does not propagate after unblocking in leaderboard


VRService:
Small issue, but was discovered yesterday whilst testing with Rootie_Groot

![Demo](https://image.prntscr.com/image/mUXZBKAxTYmLTxiqhc4zbg.png)

Repro steps:

1. Find anyone with an administrative / other custom icon that is normally awarded via getCustomPlayerIcon function
2. Block this user
3. Unblock this user

Result: The user will not have their custom badge, instead their builder club / other status will be shown.

Submitting pull request for my suggested fix

PR #1017, opened 2017-08-13
Add call to getCustomPlayerIcon function in blockStatusChanged

VRService:
See issue https://github.com/Roblox/Core-Scripts/issues/1016 for more information

(tl;dr admin icons not shown after unblock, fixes this behaviour)

I'd like to point out that, due to the nature of this issue, testing has been limited because I have no access to CoreScript functions.

### DIFF
--- a/CoreScriptsRoot/Modules/PlayerlistModule.lua
+++ b/CoreScriptsRoot/Modules/PlayerlistModule.lua
@@ -1786,12 +1786,14 @@ local blockStatusChanged = function(userId, isBlocked)
 
   for _,playerEntry in ipairs(PlayerEntries) do
     if playerEntry.Player.UserId == userId then
-      playerEntry.Frame.BGFrame.MembershipIcon.Image = getMembershipIcon(playerEntry.Player)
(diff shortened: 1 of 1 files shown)


stillunt1tled:
Sadly, I cannot say if this works or not, as it is impossible for me to test. The fake users created when you make a Local Server are technically guests (due to their negative userID's), so they cannot block people.

But looking at the changes, it *should* work.

VRService:
@stillunt1tled Yeah - I did some tests with faking block / unblock signals and it *seemed* to work properly, however an Engineer will need to check this out properly to ensure no issues

hockeychaos:
This says conflicts, but I've merged in a fix that is basically identical to this. Thanks for the pull request!

stillunt1tled:
? @hockeychaos last time I checked, this had zero conflicts. It might because you merged a ton of PR's recently. Oh well, at least it was fixed.

VRService:
Hey @hockeychaos 
I just tried this again with InceptionTime yesterday - still occurs

Example

Is this live?

PR #1024, opened 2017-08-15
Change Clamp to math.clamp

Kampfkarren:


PhoenixShay:
Not sure why this is needed. Probably a reason why they are using a custom function.

JanBerktold:
That code has likely been written before the addition of math.clamp to ROBLOX lua. Less code to maintain is usually better.

kyle-emmerich:
Yeah, this code just predates me adding math.clamp. Unfortunately, the function signatures don't match. You'll want to rearrange them. The old Clamp function here takes min, max, x, but the math.clamp function takes x, min, max.

boynedmaster:
You're right. I'll fix that when I'm at a computer and I remember.

boynedmaster:
There, that should fix it.

cliffchapmanrbx:

PR #1026, opened 2017-08-15
Add "ReportPlayer" SetCore functionality

boynedmaster:
Right now, game developers can only tell players to report bad users through a few means. The first is by reporting the player through the settings hub (Escape menu). This can be a problem with games with servers of many players, where it is difficult to scroll through every single name until you find the player you want to report. The second is by clicking on the player's name in the player list and clicking Report Abuse. Not only does this have the same problems of scrolling through to find the user's name, a lot of games disable this list. The last method is by going to the Roblox website to report them, which is inconvenient for a player to have to go through.

This PR has a solution for the problem of inconvenient reporting methods. By letting a developer report a player through SetCore, they can add a button to their own player lists, next to parties from users, and more. Developers who abuse this feature will have punitive actions taken against them similar to those who abuse the MessagePosted chat connection.

# API

StarterGui:SetCore("ReportPlayer", table<Player, string, Ref<string>> configTable)

configTable has the format of {Player playerToReport, string reportReason, Ref<string> optionalMessage}.

### DIFF
--- a/CoreScriptsRoot/CoreScripts/ReportPlayer.lua
+++ b/CoreScriptsRoot/CoreScripts/ReportPlayer.lua
@@ -1,0 +1,51 @@
+--[[
+		Filename: ReportPlayer.lua
+		Written by: boynedmaster/Kampfkarren
+		Version 1.0
+		Description: Adds ReportPlayer SetCore function.
+        Notes:       This basically just acts as a wrapper for Players:ReportAbuse, but in a context level normal LocalScripts can access.
+                     It is a separate file rather than inside a similar script, such as ReportAbuseMenu.lua, to ensure the SetCore functionality exists on startup.
+                     Why you would want to use it on startup I have no clue, but TheGamer101 said on the developer forum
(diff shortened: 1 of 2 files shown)


ConorGriffin37:
We could not allow a setcore method like this, because of the potential for developers to abuse it. The only way we could do something like this is if we allowed developers to prompt players asking do they want to report a player, similar to the prompt block and prompt friend. 

This would require a lot of internally discussion on if we wanted a feature like this though, so it is not the type of pull request we could accept here.

boynedmaster:
I understand, but shouldn't the same logic apply to MessagePosted? What stops that from being abused?

ConorGriffin37:
MessagePosted allows a level of abuse that is already pretty much possible when a developer uses a fully custom chat. They are able to make it look like a user said something they did not say to other users and cause them to be reported. 
This would allow perhaps a higher level of abuse because a developer could completely spam the report abuse queue with fake reports, causing it to be difficult for moderators to do their jobs.

boynedmaster:
Okay. Because of the rules on community PRs, I am unable to create the prompt myself so I will just wait for the internal verdict.

PR #1027, opened 2017-08-16
Change use of deprecated property "archivable"

Kampfkarren:
Unrelated, but does anyone know why properties like Pitch, Looped, etc are set *after* the Parent?


boynedmaster:
I am keeping this up although I have made a second PR completely replacing the current sound system just in case that gets denied for whatever reason.

cliffchapmanrbx:

Issue #1028, opened 2017-08-16
Sound StarterCharacterScript is needlessly insecure


boynedmaster:
For those who don't know, there is a property in the Sound object called "Playing", which can be used to toggle whether a sound is playing or not, as the name suggests. The reason I bring this up is because this Property is different from almost every other Roblox property in that it replicates to the server, that is it bypasses non-Experimental Mode. checks 

This is a problem and has caused any game unaware of it to have an exploit where a hacker can repeatedly play every sound in the game over and over again, regardless of whether Experimental Mode is on or off. This has caused many developers, including me, to write our own sound systems that create sounds on the client rather than having them on the server at all. 

The Sound StarterCharacterScript responsible for creating the sounds of a character (walking, jumping, dying, etc) problematically creates the sounds of the character on the server. This means any unaware developer will have this exploit used on their games with every sound a character could have be repeatedly played on loop. There is no reason for the sounds to be created on the server. LocalSound should be completely responsible for the playing and creating of sounds without any involvement of the server. This will solve the issue of the exploit (for character sounds at least) because all the sounds are created on the client, not on the server.

I can create this functionality as I already have done for my own game, but the main problem is backwards compatibility. I can assume some games rely on the sounds being created on the server, so the more secure character sound system will have to be enabled rather than be the default. Because of the limitations of CoreScripts (in this case, CoreScripts cannot create custom properties), it will have to be a dirty way to implement rather than a property as it should be. When I create my PR, if it gets approved I ask that it not be merged until Roblox creates a property on either SoundService or more accurately StarterPlayer to enable it. I think the way I am going to have it enabled is it will check if you have an object named "UseNewSoundSystem" in SoundService.

Jody7:
I believe this is resolved by setting Soundscape to "RespectFilteringEnabled"

boynedmaster:
That's a feature? I didn't know that. My bad.

PR #1029, opened 2017-08-16
Created new, more secure sound system

boynedmaster:
Solution for #1028.

You can test it out here: https://www.roblox.com/games/983306953/sound-test.

To use the new sound system, create an object (doesn't matter what type) in SoundService named UseNewSoundSystem.

Again, as I said in the original issue, this should *not* be pushed until Roblox creates a property for it (or if it is confirmed they won't). Having to create an object instead is a hacky method of doing so.

Not sure why the commits are showing up as being made by Unknown, I'm just going to assume I somehow screwed up my git commands.


boynedmaster:
Not sure what that conflict is about, can anyone inform me?

stillunt1tled:
@boynedmaster conflicts generally happen when another PR/change is pushed and this PR does not include that change. Can be solved by pulling the current commit to your fork. (in command-line you can type `git pull`)

boynedmaster:
There, I believe I fixed it.

hockeychaos:
unfortunately we only look for bug fixes in pull requests from the community. A full rewrite of a system or new features will not be accepted.

PR #1034, opened 2017-08-29
Add an easy way to replace moveFunc

einsteinK:
Add a getter and setter for moveFunc, for easy modifying move behaviour without having to edit the MasterControl script

### DIFF
--- a/PlayerScripts/StarterPlayerScripts/ControlScript/MasterControl.lua
+++ b/PlayerScripts/StarterPlayerScripts/ControlScript/MasterControl.lua
@@ -163,6 +163,13 @@ function MasterControl:GetMoveVector()
 	return moveValue
 end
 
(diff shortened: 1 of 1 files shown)


blobbyblob:
This is a feature request on my behalf. I want to use standard controls because users are already familiar with them, and it's easier for me to stay up-to-date on mobile/console/PC. However, I need the custom logic of only allowing the player to move on a grid.

I work around the lack of such a feature by listening to changes on Humanoid.MoveDirection & setting the humanoid's TargetPoint. I have little faith in this continuing to work going forward and also want to mitigate the risk of two scripts doing this logic simultaneously (seems like an easy avenue for an infinite loop which would crash the client).

This approach seems like an unintrusive way to allow users to provide some customization without having to edit the script manually (and lose future updates).

cliffchapmanrbx:

PR #1036, opened 2017-09-15
Prevent server scripts from timing out

hunte922:
Very often I get cheaters in my game who are able to break the server scripts by lagging out the server due to excessive RemoteEvent and RemoteFunction requests. This fixes that issue.


boynedmaster:
is there any reason most of the methods in ScriptContext are protected? this seems like one of the many functions that really has no need to be so secret.

PR #1037, opened 2017-10-08
Clean up Sound.server

stillunt1tled:
The formatting on it is just.. bleh

(aka remove the really ugly and confusing tabs)


stillunt1tled:
***screw it, probably just going to get denied anyway***

PR #1045, opened 2018-01-02
Prevent infinite yield in FreeCamera

suufi:
Infinite yield possible so let's put a timeout. :stuck_out_tounge:


Kampfkarren:
This'll just make the script error if they don't exist in the 5 seconds. Why care about the infinite yield? Better than errors.

suufi:
I mean why won't they exist?

Kampfkarren:
Why worry about an infinite yield then?

cliffchapmanrbx:

PR #1046, opened 2018-01-08
fix for deprecated PromptProductPurchaseFinished event and #1006

ZachCurtis:
extension of #1006 but fixing incorrect 2nd parameter, passing `PurchaseData.ProductId` instead of `PurchaseData.AssetId`

Deprecated event documentation can be found [here](http://wiki.roblox.com/index.php?title=API:Class/MarketplaceService/PromptProductPurchaseFinished)

### DIFF
--- a/CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
+++ b/CoreScriptsRoot/CoreScripts/PurchasePromptScript2.lua
@@ -978,7 +978,7 @@ local function onPromptEnded(isSuccess)
 
 	closePurchaseDialog()
 	if IsPurchasingConsumable then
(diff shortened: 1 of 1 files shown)


Kampfkarren:
Don't do this. PromptProductPurchaseFinished _is_ deprecated, but for illegitimate reasons. PromptProductPurchasedFinished IS NOT the same as PromptPurchaseFinished. In fact, it is the only way to see if the player denies a product purchase. This will break many games.

stillunt1tled:
I strongly agree with @Kampfkarren, this would break basically **every single game** that uses Game Passes and properly handles their in-game purchase.

Issue #1047, opened 2018-02-13
Whisper issue


loser-RBLX:
It has come to my attention that you cannot whisper users with spaces in their name.

edit evidence:
https://imgur.com/xY5LRdU

stillunt1tled:
Users should **never** have spaces in their names.

If it *appears* they have a space, its a just a underscore (`_`), and as visible in your image, it *is* a underscore.

Spaces have not been allowed in names since like 2009.

loser-RBLX:
Nevertheless it's still an issue that is rather agitating if you do so happen to meet a user with a space in their username despite the probability.

cliffchapmanrbx:
Hi there, thank you for your feedback. We're going to be closing this repository as we've chosed [the Roblox Dev Forum](https://devforum.roblox.com/) as our venue for developer feedback. Please feel free to migrate your feedback to a DevForum post if you feel it is still relevant.

PR #1050, opened 2018-03-10
fix for #1047

ZachCurtis:
Makes the conditional logic structure horribly ugly, but I know how the rules related to large edits apply here. #1047 is kind of a major bug; I've seen users with spaces in their name still playing recently.


cliffchapmanrbx:

Issue #1043, opened 2017-12-27
Chat Signals are only lowercase methods, which is out of standard with Roblox's implementations


Quenty:
https://github.com/Roblox/Core-Scripts/blob/899ad4877fb2e953b5251f2061a50c1f223d8629/CoreScriptsRoot/Modules/ChatUtil.lua#L17

Unlike Roblox signals, where `connect` is depreciated, and `Connect` is preferred, the ChatUtil only implements lowercase `:connect()`

boynedmaster:
Same with :wait() and :Wait().

stillunt1tled:
I'm assuming this is also true with :fire()?

cliffchapmanrbx:
Hi there, thank you for your feedback. We're going to be closing this repository as we've chosed [the Roblox Dev Forum](https://devforum.roblox.com/) as our venue for developer feedback. Please feel free to migrate your feedback to a DevForum post if you feel it is still relevant.

PR #1051, opened 2018-03-17
[#1043] Deprecate lower camel case method names

ghost:
Resolves issue #1043 by renaming the methods to reflect the method names of BindableEvent.

`Signal:fire() -> Signal:Fire()`
`Signal:connect() -> Signal:Connect()`
`Signal:wait() -> Signal:Wait()`

The old methods *still* exist for compatibility, but are deprecated


Kampfkarren:
You can clean up your code by instead of doing...

```
function sig:connect(f)
	self:Connect(f)
end
```

...just do

`sig.connect = sig.Connect`

stillunt1tled:
@Kampfkarren People using the old API did `sig:connect()`, not `sig.connect()`, so this would not work.

Kampfkarren:
@stillunt1tled `x:y(...)` is the exact same as `x.y(x, ...)`.

Thus, if you have this code

```
local X = {}

function X:A()
    print("hello")
end

X.B = X.A
```

Then attempting to call `X:B()` will do the exact same thing as `X:A()`. This is because `X:B()` will now be equivalent to `X.A(X.A)`.

My point is, `sig.connect = sig.Connect` is the exact same as what you're doing, just more efficient and readable.

stillunt1tled:
@Kampfkarren while I will take your suggestion, if this is done, `self` will never be passed to the method (not that its used in this script, I just concerned about future implementations that use `self`)

J0shhT:
@stillunt1tled The colon operator automatically passes self as a parameter to the function. That's the whole point of the colon operator in the first place.

https://www.lua.org/manual/5.1/manual.html#2.5.8

"A call v:name(args) is syntactic sugar for v.name(v,args), except that v is evaluated only once."

stillunt1tled:
^ this is why I don't contribute to this repository.

Thats the point. If you dont have the `:` operator, self wont be passed. I'm saying if they use it in a future version of this script, all of the

Kampfkarren:
Er, no. self is automatically passed.

https://repl.it/repls/IdealisticWebbedLogins

cliffchapmanrbx:

PR #1052, opened 2018-03-19
Update to latest internal version of core scripts.

tiffany352:
It's been 7 months, oops.


Kampfkarren:
YES! Does this mean that we can start contributing again, like I implied in my dev forum thread?

hockeychaos:
we are evaluation our options as to how to integrate this with our internal source control. Generally it is hard to sync between the two so we want to make sure this is an easy process

Kampfkarren:
Is the internal source control system developed in house? Is it not something normal like git or mercurial?

tiffany352:
We use Perforce internally for the engine, client, studio, and any resources they require. https://en.wikipedia.org/wiki/Perforce

Kampfkarren:
I see. I have no knowledge of Perforce, but it may be worth your while to look into Git Fusion. https://www.perforce.com/video-tutorials/git-fusion-overview. I am very happy to hear that community contribution to core scripts is at the very least back on the table.

hockeychaos:
we use Perforce, so really it is a questions of figuring out how to integrate with git

PR #1060, opened 2018-07-13
Removed the slower clamp function.

howmanysmall:
My recommendation is to change all the clamp functions to the built-in math.clamp now that it exists.


idiomic:
Looks like a duplicate of https://github.com/Roblox/Core-Scripts/pull/1024

PR #1062, opened 2018-09-03
Update FreeCamera.lua

howmanysmall:
Optimizations and various fixes.

 - Removed the useless and slow clamp function for the faster `math.clamp`.
 - Replaced the single `math.abs` for a faster version of it that doesn't involve a function call.
 - Formatting changed.
 - Changed lowercase things like `:wait` and `.x` to their uppercase variants `:Wait` and `.X`.


cliffchapmanrbx:
Hi there, thank you for your feedback. We're going to be closing this repository as we've chosed [the Roblox Dev Forum](https://devforum.roblox.com/) as our venue for developer feedback. Please feel free to migrate your feedback to a DevForum post if you feel it is still relevant.

